PR #38, opened 2024-02-18
Add: activities-set-frame-name - make setting frame name optional

### LINKED ISSUE #33: Optionally do not `set-frame-name`
Many people customize the frame title to include other information (like full path).  It would be great if `activities` had an option _not_ to set the frame name.  Then you can always include an `:eval` in `frame-title-format` with `(activities-name-for (activities-current))`.

jdtsmith:
Fixes #33.  This can be used with a custom `frame-title-format`, e.g.:

```elisp
(setq frame-title-format
	`((:eval (when-let ((activities-mode)
			    (cur (activities-current)))
		   (concat (activities-name-for cur) " • ")))
	  ,@frame-title-format))
```

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -279,8 +279,8 @@ prefixes will be added automatically."
                  (const :tag "Current project's name" activities--project-name)
                  (function-item :tag "Other function")))
 
-(defcustom activities-inhibit-frame-name nil
-  "Whether to inibit setting the frame name to the current activity name.
+(defcustom activities-set-frame-name t
+  "Set frame name after switching activity.
 Only applies when `activities-tabs-mode' is disabled."
   :type 'boolean)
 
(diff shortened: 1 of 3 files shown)


activities.el
```diff
@@ -279,6 +279,11 @@ prefixes will be added automatically."
                  (const :tag "Current project's name" activities--project-name)
                  (function-item :tag "Other function")))
 
+(defcustom activities-inhibit-frame-name nil
```
alphapapa, line 282: Naming an option in negative terms creates a "double-negative problem," where the user has to think in terms of negating a negative to understand what it does.  If we name it simply `-set-frame-name`, it will be clearer.

activities.el
```diff
@@ -279,6 +279,11 @@ prefixes will be added automatically."
...
+  "Whether to inibit setting the frame name to the current activity name.
```
alphapapa, line 283: Minor typo.  :)  We can just say, "Set frame name after switching activity."

alphapapa (changes requested):
Thanks.  I think a better long-term solution would be to add an after-switch hook.  But in the meantime, an option will solve this problem.

alphapapa:
Thanks.

PR #50, opened 2024-03-14
Use 'ignore' instead of an equivalent 'lambda' form

eshelyaron:
Hello Adam,

This is just a small simplification that came to mind while reading through some of the code of this very nice package :)


alphapapa:
Hi Eshel,

Thanks, I'll merge this for the next version.

PR #57, opened 2024-03-26
Add a configurable width for the time columns in `activities-list`

johnhamelink:
This ensures that the columns are not truncated by default.

Screenshot 2024-03-26 at 18 50 36

### DIFF
--- a/activities-list.el
+++ b/activities-list.el
@@ -36,6 +36,11 @@
   "Time format for `activities-list' buffer."
   :type 'string)
 
+(defcustom activities-list-time-column-width 20
+  "`activities-list' column-width for columns which display time."
+  :type '(choice (integer :tag "Number of chars")
+                 (string :tag "Width String (See info node `(vtable)Top' for more information.)")))
+
 (defmacro activities-list-command (command)
   "Expand to a lambda that applies its args to COMMAND and reverts the list buffer."
   `(lambda (&rest args)
@@ -68,12 +73,14 @@
                      (pcase-let (((cl-struct activities-activity last) activity))
                        (when last
                          (map-elt (activities-activity-state-etc last) 'time))))
+           :width ,activities-list-time-column-width
            :formatter activities-list--format-time)
          ( :name "Default saved"
            :getter (lambda (activity _table)
                      (pcase-let (((cl-struct activities-activity default) activity))
                        (when default
                          (map-elt (activities-activity-state-etc default) 'time))))
+           :width ,activities-list-time-column-width
            :formatter activities-list--format-time))
        :objects-function (lambda ()
                            (map-values activities-activities))


alphapapa:
Hi John,

Thanks, but I don't understand why this is necessary.  Those columns aren't truncated on my system.  Would you help me understand the problem?

johnhamelink:
Hi 😄 

> Would you help me understand the problem?

Certainly. When I run `activities-list` from this repo, the table looks like:

Screenshot 2024-03-26 at 20 33 51

`describe-char` on a timestamp shows the following:

```
             position: 41 of 74 (54%), column: 9
            character: 0 (displayed as 0) (codepoint 48, #o60, #x30)
              charset: ascii (ASCII (ISO646 IRV))
code point in charset: 0x30
               script: latin
               syntax: w 	which means: word
             category: .:Base, a:ASCII, l:Latin, r:Roman
             to input: type "C-x 8 RET 30" or "C-x 8 RET DIGIT ZERO"
          buffer code: #x30
            file code: #x30 (encoded by coding system utf-8)
              display: by this font (glyph code):
    mac-ct:-*-Iosevka-regular-normal-normal-*-16-*-*-*-m-0-iso10646-1 (#x13)

Character code properties: customize what to show
  name: DIGIT ZERO
  general-category: Nd (Number, Decimal Digit)
  decomposition: (48) ('0')
```

While I can extend the columns manually by tapping `}`, it's annoying to have to do it every time I open the view, so I configured the width of the column.

alphapapa:
I don't know.  It seems to work correctly for me.  Here's a screenshot:

example

Like in yours, the header line and name column use variable pitch, and the dates use monospace.

AFAIK if those columns are being truncated, it would seem to point to a bug somewhere, because their widths should be calculated automatically, and none of them should be truncated unless the table is wider than the window.

But I would recommend, yes, trying a different font to see if it makes any difference.

johnhamelink:
Yes, so the culprit seems to be the following:

```emacs-lisp
(set-face-attribute 'variable-pitch nil
                      :font "ETBembo"
                      :weight 'thin
                      :height 150)
```

Disabling this variable-pitch font seems to fix the truncation issue for me.

alphapapa:
Ok, I guess that's a bug in either `vtable` or another part of Emacs that calculates string widths, so it should probably be filed as an Emacs bug.  Bonus points for a minimal reproducible test case.  :)

alphapapa:
@johnhamelink If you do file the bug upstream, please link it here for future reference.

johnhamelink:
@alphapapa I wrote a minimal example, but the issue is not appearing for me anymore - instead I'm experiencing what I think is ligature-related inconsistencies which I understand is a known problem.

For reference the example I wrote is as follows:

```emacs-lisp
(require 'vtable)

(defun build-sample-vtable ()
  "Build an example VTable."

  (make-vtable
   :objects '(("" "Ruby" "2024-03-27 13:04:46" "2024-03-27 13:04:46")
              ("" "Emacs" "2024-03-27 13:04:46" "2024-03-27 13:04:46")
              ("" "Exercism [Elisp]" "2024-03-27 13:04:46" "2024-03-27 13:04:46")
              ("" "Nix" "2024-03-27 13:04:46" "2024-03-27 13:04:46"))
   :columns
   `((:name "Active" :primary descend)
     (:name "Name")
     (:name "Last saved")
     (:name "Default saved"))
   :sort-by '((2 . descend) (0 . descend))))

(set-face-attribute 'fixed-pitch nil
                        :font "Iosevka Comfy Wide Motion Fixed"
                        :height 160)

(set-face-attribute 'default nil
                        :font "Iosevka"
                        :weight 'regular
                        :height 160)

(switch-to-buffer (generate-new-buffer "without-variable-font"))
(build-sample-vtable)

(with-selected-frame (make-frame)
  (switch-to-buffer (generate-new-buffer "with-variable-font"))
  (face-remap-add-relative
   'variable-pitch  nil
   '(:font "ETBembo" :weight 'thin :height 120))
  (build-sample-vtable))
```

Which produces the following:

Screenshot 2024-03-28 at 13 07 58

I've sent a bug report (70055) which can be found here: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=70055

Thanks!

alphapapa:
Thanks.  I can imagine that ligatures might affect this kind of problem.  Please be sure to mention that in the bug report.

FWIW, it looks like the buffer names are reversed, i.e. "with-variable-font' appears to have the fixed-width font in use.

PR #77, opened 2024-04-13
Fix: (activities-list-command) Use vtable-revert-command

josephmturner:
It seems to be necessary to clear the vtable cache in order for the listings to properly refresh.


alphapapa:
Thanks.

PR #79, opened 2024-04-14
Fix: Use default tab-bar-tab-face-function

### LINKED ISSUE #76: Activities tabs should resize according to `tab-bar-auto-width`
Activities tabs are not resized even when the `tab-bar-auto-width` option is non-nil.

```
;; Create a bunch of activities, see the tab bar wrap 
(progn
  (setopt tab-bar-auto-width t)
  (activities-tabs-mode 1)
  (dotimes (i 10)
    (activities-new (format "test-long-name-%s" i))))

;; Clean up
(progn
  (dotimes (i 10)
    (let* ((name (format "test-long-name-%s" i))
           (activity (map-elt activities-activities name)))
      (ignore-errors
        (when (activities-activity-active-p activity)
          (activities-close activity)))
      (setf activities-activities (map-delete activities-activities name)))))
```

Internally, the `tab-bar-auto-width` function determines that a tab should be resized if its face is one of `tab-bar-auto-width-faces`.  However, `activities-tabs--tab-bar-tab-face-function` (say that even once fast... 😉) sets the activities tab face to `(:inherit (activities-tabs ,face))`, so the activities tabs are never resized.

josephmturner:
This resolves a problem where activities tabs were not resized when the tab-bar-auto-width option is non-nil.

Internally, the tab-bar-auto-width function determines that a tab should be resized if its face is one of tab-bar-auto-width-faces. However, activities-tabs--tab-bar-tab-face-function sets the activities tab face to (:inherit (activities-tabs ,face)), so the activities tabs were never resized.

By using the default face, the memq check works as intended.

Resolves #76.

### DIFF
--- a/activities-tabs.el
+++ b/activities-tabs.el
@@ -202,10 +202,11 @@ Selects its tab, making one if needed.  Its state is not changed."
 If TAB represents an activity, face `activities-tabs' is added as
 inherited."
   ;; TODO: Propose a tab-bar equivalent of `tab-line-tab-face-functions'.
-  (let ((face (funcall activities-tabs-tab-bar-tab-face-function-original tab)))
-    (if (activities-tabs--tab-parameter 'activity tab)
-        `(:inherit (activities-tabs ,face))
-      face)))
+  ;; (let ((face (funcall activities-tabs-tab-bar-tab-face-function-original tab)))
+  ;;   (if (activities-tabs--tab-parameter 'activity tab)
+  ;;       `(:inherit (activities-tabs ,face))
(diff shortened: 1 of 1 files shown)


josephmturner:
This workaround should not be necessary if https://yhetil.org/emacs-bugs/87le2kaloz.fsf@breatheoutbreathe.in/T/#u is merged.

alphapapa:
I guess we won't add this workaround, but what do you think?

josephmturner:
Yes, I agree.

PR #81, opened 2024-04-14
Change: Remove defcustom activities-tabs-before-resume-functions

josephmturner:
Was `activities-tabs-before-resume-functions` a holdover from early development?  Since we already have `activities-before-resume-functions`, why do we need it?


alphapapa:
I don't remember why I added this, but I'm not sure we should remove it.  A user might want to use it to do something when resuming an activity when using tabs mode.  Or am I missing something?

josephmturner:
I see.  I think it's fine to leave it in!

PR #83, opened 2024-04-20
annotations and sorting for activities-completing-read

jdtsmith:
(updated) This provides custom annotation and sorting functions for `activities-completing-read`. It shows the number of buffers (and files), for the last or default state, a magit-style color-coded _activity age_, and flags for being active or having a modified set of buffers. 

This is shown below with `vertico` but it should support any UI which handles annotation-functions (most do).   For theme friendliness, it creates no new faces, instead ~borrowing the `vc-annotate` color palette~ building its own color palette (along with `success` and `warning` faces).  Happy to take any input/ideas.

image

### DIFF
--- a/README.org
+++ b/README.org
@@ -164,7 +164,11 @@ The sort order when completing activities can be configured using ~activities-so
 
 ** v0.8-pre
 
-Nothing new yet.
+*Additions*
++ New completion interface with rich annotations. (See #83.)
+
+*Changes*
++ During save, the ~time~ slot for an activity remains unchanged unless its buffers or files differ from their last saved state. (See #83.)
 
 ** v0.7.2
 
--- a/README.org
+++ b/README.org
@@ -135,6 +135,10 @@ Key bindings are, as always, ultimately up to the user.  However, in [[Configura
 
 When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored when a new activity is made.  This allows the command ~bookmark-jump~ (~C-x r b~) to be used to resume an activity (helping to universalize the bookmark system).
 
+** Annotation Info
+
+When selecting an activity in the minibuffer using an interface that supports /annotations/, additional information is shown: the number of buffers and files the activity contains, a color-coded age since the activity was last updated, and a final ~*~ if the activity's list of buffers has been modified from its default state.
+
 * FAQ
 
 + How is this different from [[https://github.com/alphapapa/burly.el][Burly.el]] or [[https://github.com/alphapapa/bufler.el/][Bufler.el]]? :: Burly is a well-polished tool for restoring window and frame configurations, which could be considered an incubator for some of the ideas furthered here.  Bufler's ~bufler-workspace~ library uses Burly to provide some similar functionality, which is at an exploratory stage.  ~activities~ hopes to provide a longer-term solution more suitable for integration into Emacs.
--- a/README.org
+++ b/README.org
@@ -137,12 +137,14 @@ When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored 
 
 ** Completion
 
-When selecting an activity in the mini-buffer, annotations are sorted active first, then by most recent modification age.  If you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
+When selecting an activity in the mini-buffer, if you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
 
 - an `@` symbol, for active activities,
 - the number of buffers and files the activity contains,
 - a color-coded age since the activity was last updated, and
 - a final ~*~ if the activity's list of buffers and files has been modified from its default state.
+
+The sort order when completing activities can be configured using ~activities-sort-by~.
 
 * FAQ
 
--- a/README.org
+++ b/README.org
@@ -137,12 +137,12 @@ When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored 
 
 ** Completion
 
-When selecting an activity in the mini-buffer, if you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
+When selecting an activity in the minibuffer, if you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
 
-- an `@` symbol, for active activities,
-- the number of buffers and files the activity contains,
-- a color-coded age since the activity was last updated, and
-- a final ~*~ if the activity's list of buffers and files has been modified from its default state.
+- An ~@~ symbol for active activities,
+- The number of buffers and files the activity contains,
+- A color-coded age (elapsed time since the activity was last updated), and
+- A final ~*~ if the activity's list of buffers and files has been modified from its default state.
 
 The sort order when completing activities can be configured using ~activities-sort-by~.
 
@@ -165,10 +165,10 @@ The sort order when completing activities can be configured using ~activities-so
 ** v0.8-pre
 
 *Additions*
-+ New completion interface with rich annotations. (See #83.)
++ New completion interface with rich annotations.  ([[https://github.com/alphapapa/activity.el/pull/83][#83]].  Thanks to [[https://github.com/jdtsmith][JD Smith]].)
 
 *Changes*
-+ During save, the ~time~ slot for an activity remains unchanged unless its buffers or files differ from their last saved state. (See #83.)
++ During save, the ~time~ slot for an activity remains unchanged unless its buffers or files differ from their last saved state.  ([[https://github.com/alphapapa/activity.el/pull/83][#83]].  Thanks to [[https://github.com/jdtsmith][JD Smith]].)
 
 ** v0.7.2
 
--- a/README.org
+++ b/README.org
@@ -135,9 +135,14 @@ Key bindings are, as always, ultimately up to the user.  However, in [[Configura
 
 When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored when a new activity is made.  This allows the command ~bookmark-jump~ (~C-x r b~) to be used to resume an activity (helping to universalize the bookmark system).
 
-** Annotation Info
+** Annotation and Sort
 
-When selecting an activity in the minibuffer using an interface that supports /annotations/, additional information is shown: the number of buffers and files the activity contains, a color-coded age since the activity was last updated, and a final ~*~ if the activity's list of buffers has been modified from its default state.
+When selecting an activity in the mini-buffer, annotations are sorted active first, then by most recent modification age.  If you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
+
+- an `@` symbol, for active activities,
+- the number of buffers and files the activity contains,
+- a color-coded age since the activity was last updated, and
+- a final ~*~ if the activity's list of buffers and files has been modified from its default state.
 
 * FAQ
 
--- a/README.org
+++ b/README.org
@@ -135,7 +135,7 @@ Key bindings are, as always, ultimately up to the user.  However, in [[Configura
 
 When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored when a new activity is made.  This allows the command ~bookmark-jump~ (~C-x r b~) to be used to resume an activity (helping to universalize the bookmark system).
 
-** Annotation and Sort
+** Completion
 
 When selecting an activity in the mini-buffer, annotations are sorted active first, then by most recent modification age.  If you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
 
--- a/activities.el
+++ b/activities.el
@@ -341,7 +341,9 @@ applied.  A custom predicate function may also be set.  It should
 take two arguments, both activity names (strings), and return
 non-nil if the first activity should sort before the second."
   :type `(choice (const :tag "No sorting" nil)
-		 (const :tag "Active state and age" ,#'activities-sort-by-active-age)
+		 (function-item :tag "Active state and age"
+				:doc "Sort by active state and age."
+				,#'activities-sort-by-active-age)
 		 (function :tag "Custom predicate")))
 
 (defcustom activities-annotation-colors '("blue" "red" 0.65)
--- a/activities.el
+++ b/activities.el
@@ -850,11 +850,9 @@ Adapted from `magit--age'."
   (cl-loop for (_name . act) in activities maximize
 	   (cl-loop
 	    for type in '(default last)
-	    for func = (intern (format "activities-activity-%s" type))
-	    maximize
-	    (float-time
-	     (time-since
-	      (map-elt (activities-activity-state-etc (funcall func act)) 'time))))))
+ 	    for state = (cl-struct-slot-value 'activities-activity type act)
+	    for etc = (activities-activity-state-etc state)
+	    maximize (float-time (time-since (map-elt etc 'time))))))
 
 (defun activities-annotate (max-age oldest-possible)
   "Return an activity annotation function.
@@ -864,8 +862,7 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
     (when-let ((activity (map-elt activities-activities name)))
       (let (data (age-len 14))
 	(dolist (type '(last default))
-	  (let* ((func (intern (format "activities-activity-%s" type)))
-		 (state (funcall func activity))
+	  (let* ((state (cl-struct-slot-value 'activities-activity type activity))
 		 (time (map-elt (activities-activity-state-etc state) 'time))
 		 (window-state (activities-activity-state-window-state state))
 		 (buffers (window-state-buffers window-state))
--- a/activities.el
+++ b/activities.el
@@ -535,8 +535,10 @@ form (BUFFER . FILE) assocaited with the activity."
   (activities--map-window-state-leafs
    (activities-activity-state-window-state state)
    (lambda (leaf)
-     (let ((rec (map-nested-elt (cdr leaf) '(parameters activities-buffer))))
-       (cons (activities-buffer-name rec) (activities-buffer-filename rec))))))
+     (let ((buffer-rec (map-nested-elt (cdr leaf)
+				       '(parameters activities-buffer))))
+       (cons (activities-buffer-name buffer-rec)
+	     (activities-buffer-filename buffer-rec))))))
 
 (defun activities--buffers-and-files-differ-p (bf1 bf2)
   "Return non-nil if BF1 and BF2 are not the same set of files or buffers.
--- a/activities.el
+++ b/activities.el
@@ -950,19 +950,6 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
 		 (activep-b (activities-activity-active-p activity-b)))
 	    (or (and activep-a (not activep-b)) (time-less-p time-b time-a))))))
 
-(defun activities--completion-table (activities)
-  "Return a completion table function to complete ACTIVITIES."
-  (let ((names (activities-names activities)))
-    (lambda (str pred action)
-      (if (eq action 'metadata)
-	  `(metadata
-	    (annotation-function . ,(activities--annotate (activities--oldest-age activities)
-							 (vc-annotate-oldest-in-map
-							  vc-annotate-color-map)))
-	    ,@(when activities-sort-function
-		`(,(cons 'display-sort-function activities-sort-function))))
-	(complete-with-action action names str pred)))))
-
 (cl-defun activities-completing-read
     (&key (activities activities-activities)
           (default (when (activities-current)
@@ -971,9 +958,19 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
   "Return an activity read with completion from ACTIVITIES.
 PROMPT is passed to `completing-read' by way of `format-prompt',
 which see, with DEFAULT."
-  (let* ((prompt (format-prompt prompt default))
-         (name (completing-read prompt (activities--completion-table activities) nil t nil
-				'activities-completing-read-history default)))
+  (let* ((names (activities-names activities))
+	 (table (lambda (str pred action)
+		  (if (eq action 'metadata)
+		      `(metadata
+			( annotation-function .
+			  ,(activities--annotate (activities--oldest-age activities)
+						 (vc-annotate-oldest-in-map
+						  vc-annotate-color-map)))
+			,@(when activities-sort-function
+			    `(,(cons 'display-sort-function activities-sort-function))))
+		    (complete-with-action action names str pred))))
+	 (prompt (format-prompt prompt default))
+         (name (completing-read prompt table nil t nil 'activities-completing-read-history default)))
     (or (map-elt activities-activities name)
         (make-activities-activity :name name))))
 
--- a/activities.el
+++ b/activities.el
@@ -856,9 +856,10 @@ Adapted from `magit--age'."
 	     (time-since
 	      (map-elt (activities-activity-state-etc (funcall func act)) 'time))))))
 
-(defun activities-annotate (max-age)
+(defun activities-annotate (max-age oldest-possible)
   "Return an activity annotation function.
-MAX-AGE is the maximum age of any activity in seconds."
+MAX-AGE is the maximum age of any activity in seconds.
+OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
   (lambda (name)
     (when-let ((activity (map-elt activities-activities name)))
       (let (data)
@@ -868,23 +869,22 @@ MAX-AGE is the maximum age of any activity in seconds."
 		 (time (map-elt (activities-activity-state-etc state) 'time))
 		 (window-state (activities-activity-state-window-state state))
 		 (buffers (window-state-buffers window-state))
-		 (files (activities--map-window-state-leafs window-state
-			 (lambda (l)
-			   (bookmark-get-filename
-			    (activities-buffer-bookmark
-			     (map-nested-elt (cdr l)
-			      '(parameters activities-buffer))))))))
+		 (files ( activities--map-window-state-leafs window-state
+			  (lambda (l)
+			    (bookmark-get-filename
+			     (activities-buffer-bookmark
+			      (map-nested-elt (cdr l)
+					      '(parameters activities-buffer))))))))
 	    (setf (alist-get type data)
 		  (list :bufcnt (length buffers)
 			:filecnt (length (delq nil files))
 			:time (float-time (time-since time))))))
 	(cl-labels ((data-el (&rest keys) (map-nested-elt data keys)))
-	  (let* ((oldest (vc-annotate-oldest-in-map vc-annotate-color-map))
-		 (age (min (data-el 'last :time)
+	  (let* ((age (min (data-el 'last :time)
 			   (data-el 'default :time)))
 		 (ann (format "%s:%s|%s %s:%s|%s "
 			      (propertize "buffers" 'face 'bold)
-			      (propertize (format "%2d"  (data-el 'last :bufcnt))
+			      (propertize (format "%2d" (data-el 'last :bufcnt))
 					  'face 'success)
 			      (propertize (format "%-2d" (data-el 'default :bufcnt))
 					  'face 'warning)
@@ -894,7 +894,7 @@ MAX-AGE is the maximum age of any activity in seconds."
 			      (propertize (format "%-2d" (data-el 'default :filecnt))
 					  'face 'warning)))
 		 (age-color (or (cdr (vc-annotate-compcar
-				      (* (/ age max-age) oldest)
+				      (* (/ age max-age) oldest-possible)
 				      vc-annotate-color-map))
 				vc-annotate-very-old-color))
 		 (age-ann
@@ -919,7 +919,9 @@ which see, with DEFAULT."
          (names (activities-names activities))
 	 (completion-extra-properties `(:annotation-function
 					,(activities-annotate
-					  (activities--oldest-age activities))))
+					  (activities--oldest-age activities)
+					  (vc-annotate-oldest-in-map
+					   vc-annotate-color-map))))
          (name (completing-read prompt names nil t nil 'activities-completing-read-history default)))
     (or (map-elt activities-activities name)
         (make-activities-activity :name name))))
--- a/activities.el
+++ b/activities.el
@@ -846,7 +846,7 @@ Adapted from `magit--age'."
     (fn age activities--age-spec)))
 
 (defun activities--oldest-age (activities)
-  "Return the age in seconds of the olds of ACTIVITIES."
+  "Return the age in seconds of the oldest activity in ACTIVITIES."
   (cl-loop for (_name . act) in activities maximize
 	   (cl-loop
 	    for type in '(default last)
--- a/activities.el
+++ b/activities.el
@@ -864,7 +864,8 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
 		 (time (map-elt (activities-activity-state-etc state) 'time))
 		 (window-state (activities-activity-state-window-state state))
 		 (buffers (window-state-buffers window-state))
-		 (files (activities--map-window-state-leafs window-state
+		 (files (activities--map-window-state-leafs
+			 window-state
 			 (lambda (leaf)
 			   (bookmark-get-filename
 			    (activities-buffer-bookmark
--- a/activities.el
+++ b/activities.el
@@ -849,6 +849,7 @@ Adapted from `magit--age'."
 	   (cl-loop
 	    for type in '(default last)
  	    for state = (cl-struct-slot-value 'activities-activity type act)
+	    if state
 	    for etc = (activities-activity-state-etc state)
 	    maximize (float-time (time-since (map-elt etc 'time))))))
 
@@ -860,44 +861,44 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
     (when-let ((activity (map-elt activities-activities name)))
       (let (data)
 	(dolist (type '(last default))
-	  (let* ((state (cl-struct-slot-value 'activities-activity type activity))
-		 (time (map-elt (activities-activity-state-etc state) 'time))
-		 (window-state (activities-activity-state-window-state state))
-		 (buffers (window-state-buffers window-state))
-		 (files (activities--map-window-state-leafs
-			 window-state
-			 (lambda (leaf)
-			   (bookmark-get-filename
-			    (activities-buffer-bookmark
-			     (map-nested-elt (cdr leaf)
-			 		     '(parameters activities-buffer))))))))
-	    (setf (alist-get type data)
-		  (list (length buffers)
-			(length (delq nil files))
-			(float-time (time-since time))))))
-	(pcase-let* ((`(,last-bufs ,last-files ,last-time) (map-elt data 'last))
-		     (`(,default-bufs ,default-files ,default-time) (map-elt data 'default))
-		     (age (min last-time default-time))
-		     (ann (format "%s:%s|%s %s:%s|%s "
-				  (propertize "buffers" 'face 'bold)
-				  (propertize (format "%2d" last-bufs) 'face 'success)
-				  (propertize (format "%-2d" default-bufs) 'face 'warning)
-				  (propertize "files" 'face 'bold)
-				  (propertize (format "%2d" last-files) 'face 'success)
-				  (propertize (format "%-2d" default-files) 'face 'warning)))
+	  (when-let ((state (cl-struct-slot-value 'activities-activity type activity)))
+	    (let* ((time (map-elt (activities-activity-state-etc state) 'time))
+		   (window-state (activities-activity-state-window-state state))
+		   (buffers (window-state-buffers window-state))
+		   (files (activities--map-window-state-leafs
+			   window-state
+			   (lambda (leaf)
+			     (bookmark-get-filename
+			      (activities-buffer-bookmark
+			       (map-nested-elt (cdr leaf)
+			 		       '(parameters activities-buffer))))))))
+	      (setf (alist-get type data)
+		    (list (length (delq nil files))
+			  (and time (float-time (time-since time))) buffers)))))
+	(pcase-let* ((`(,last-file-cnt ,last-age ,last-bufs) (map-elt data 'last))
+		     (`(,default-file-cnt ,default-age ,default-bufs) (map-elt data 'default))
+		     (age (if last-age (min last-age default-age) default-age))
+		     (buf-cnt (length (or last-bufs default-bufs)))
+		     (file-cnt (or last-file-cnt default-file-cnt))
+		     (dirty (and last-bufs
+				 (or (/= (length last-bufs) (length default-bufs))
+				     (not (seq-set-equal-p last-bufs default-bufs)))))
+		     (ann (format "%s bufs %s files "
+				  (propertize (format "%2d" buf-cnt) 'face 'success)
+				  (propertize (format "%2d" file-cnt) 'face 'warning)))
 		     (age-color (or (cdr (vc-annotate-compcar
 					  (* (/ age max-age) oldest-possible)
 					  vc-annotate-color-map))
 				    vc-annotate-very-old-color))
 		     (age-ann (propertize
-			       (format "%15s" (apply #'format "[%d %s]"
-						     (activities--age age)))
+			       (format "%15s" (apply #'format "[%d %s]" (activities--age age)))
 			       'face `( :foreground ,age-color
-					:background ,vc-annotate-background))))
+					:background ,vc-annotate-background)))
+		     (dirty-ann (propertize (if dirty "*" " ") 'face 'bold)))
 	  (concat (propertize " " 'display
 			      `( space :align-to
-				 (- right ,(+ (length ann) (length age-ann)))))
-		  ann age-ann))))))
+				 (- right ,(+ (length ann) (length age-ann) 1))))
+		  ann age-ann dirty-ann))))))
 
 (cl-defun activities-completing-read
     (&key (activities activities-activities)
--- a/activities.el
+++ b/activities.el
@@ -863,7 +863,7 @@ Adapted from `magit--age'."
 	    for etc = (activities-activity-state-etc state)
 	    maximize (float-time (time-since (map-elt etc 'time))))))
 
-(defun activities-annotate (max-age oldest-possible)
+(defun activities--annotate (max-age oldest-possible)
   "Return an activity annotation function.
 MAX-AGE is the maximum age of any activity in seconds.
 OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
@@ -899,7 +899,6 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
 		       (dirtyp (when last-buffers-and-files
 				 (buffers-and-files-differ-p last-buffers-and-files
 							     default-buffers-and-files)))
-		       (dirty-annotation (if dirtyp (propertize "*" 'face 'bold) " "))
 		       (annotation (format "%s%s buf%s %s file%s "
 					   (if (activities-activity-active-p activity)
 					       (propertize "@" 'face 'bold) " ")
@@ -915,7 +914,8 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
 					(format "%15s" (apply #'format "[%d %s]"
 							      (activities--age age)))
 					'face `(:foreground ,age-color
-							    :background ,vc-annotate-background))))
+							    :background ,vc-annotate-background)))
+		       (dirty-annotation (if dirtyp (propertize "*" 'face 'bold) " ")))
 	    (concat (propertize " " 'display
 				`(space :align-to (- right ,(+ 1 (length annotation)
 							       (length age-annotation)))))
@@ -945,7 +945,7 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
     (lambda (str pred action)
       (if (eq action 'metadata)
 	  `(metadata
-	    (annotation-function . ,(activities-annotate (activities--oldest-age activities)
+	    (annotation-function . ,(activities--annotate (activities--oldest-age activities)
 							 (vc-annotate-oldest-in-map
 							  vc-annotate-color-map)))
 	    ,@(when activities-sort-function
--- a/activities.el
+++ b/activities.el
@@ -902,13 +902,12 @@ Abbreviate the units if ABBREV is non-nil."
 
 (defun activities--oldest-age (activities)
   "Return the age in seconds of the oldest activity in ACTIVITIES."
-  (cl-loop for (_name . act) in activities maximize
-	   (cl-loop
-	    for type in '(default last)
- 	    for state = (cl-struct-slot-value 'activities-activity type act)
-	    if state
-	    for etc = (activities-activity-state-etc state)
-	    maximize (float-time (time-since (map-elt etc 'time))))))
+  (cl-loop for (_name . activity) in activities
+	   for state = (pcase-let (((cl-struct activities-activity default last) activity))
+			 (or last default))
+	   if state
+	   for etc = (activities-activity-state-etc state)
+	   maximize (float-time (time-since (map-elt etc 'time)))))
 
 (defun activities-sort-by-active-age (names)
   "Return the list of activity NAMES sorted active first, then by age."
--- a/activities.el
+++ b/activities.el
@@ -935,8 +935,6 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
 			      (activities-activity-default activity-b)))
 		 (time-b (map-elt (activities-activity-state-etc state-b) 'time))
 		 (activep-b (activities-activity-active-p activity-b)))
-	    (message "Comparing %s %s timeA: %S timeB: %S activeA: %S activeB: %S" a b
-		     (format-time-string "%F %r" time-a) (format-time-string "%F %r" time-b) activep-a activep-b)
 	    (or (and activep-a (not activep-b)) (time-less-p time-b time-a))))))
 
 (defun activities--completion-table (activities)
--- a/activities.el
+++ b/activities.el
@@ -66,6 +66,7 @@
 (require 'map)
 (require 'persist)
 (require 'subr-x)
+(require 'vc-annotate)
 
 ;;;; Types
 
@@ -651,6 +652,23 @@ activity's name is NAME."
 	       ;; NOTE: We copy the state so as not to mutate the one in storage.
 	       (activities--bufferize-window-state (copy-tree state))))
 
+(defun activities--map-window-state-leafs (state func)
+  "Call FUNC on all the leaf nodes in window-state STATE.
+The resulting values are returned as a list."
+  (let (ret)
+    (cl-labels ((map-leafs (state func)
+		  (pcase state
+		    (`(leaf . ,_attrs)
+		     (push (funcall func state) ret))
+		    ((pred atom) state)
+		    (`(,_key . ,(pred atom)) state)
+		    ((pred list)
+		     (if-let ((leaf-pos (cl-position 'leaf state)))
+			 (push (funcall func (cl-subseq state leaf-pos)) ret)
+		       (dolist (s state) (map-leafs s func)))))))
+      (map-leafs state func))
+    (nreverse ret)))
+
 (defun activities--bufferize-window-state (state)
   "Return window state STATE with its buffers reincarnated."
   (cl-labels ((bufferize-state (state)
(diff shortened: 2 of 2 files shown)


activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+    (when-let ((activity (map-elt activities-activities name)))
+      (let (data (age-len 14))
+	(dolist (type '(last default))
+	  (let* ((func (intern (format "activities-activity-%s" type)))
+		 (state (funcall func activity))
+		 (time (map-elt (activities-activity-state-etc state) 'time))
+		 (window-state (activities-activity-state-window-state state))
+		 (buffers (window-state-buffers window-state))
+		 (files ( activities--map-window-state-leafs window-state
+			  (lambda (l)
```
alphapapa, line 871: Nitpick: Please don't use single-letter variable names.  For clarity, just use the word.

activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+	(pcase-let* ((`(,lastb ,lastf ,lastt) (map-elt data 'last))
+		     (`(,defb ,deff ,deft) (map-elt data 'default))
+		     (age (min lastt deft))
+		     (ann (format "%s:%s|%s %s:%s|%s "
+				  (propertize "buffers" 'face 'bold)
+				  (propertize (format "%2d" lastb) 'face 'success)
+				  (propertize (format "%-2d" defb) 'face 'warning)
+				  (propertize "files" 'face 'bold)
+				  (propertize (format "%2d" lastf) 'face 'success)
+				  (propertize (format "%-2d" deff) 'face 'warning)))
```
alphapapa, line 889: This looks nice, but the variable names are a bit unclear with those abbreviations.  I'd rather the code be clear and unambiguous, even if it means the formatting suffers.  So please just use the full words in the variable names.

activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+		     (age-color (or (cdr (vc-annotate-compcar
+					  (* (/ age max-age) oldest-possible)
+					  vc-annotate-color-map))
+				    vc-annotate-very-old-color))
+		     (age-ann
+		      (propertize (apply #'format "[%d %s]" (activities--age age)) 'face
+				  `( :foreground ,age-color
+				     :background ,vc-annotate-background))))
+	  (when (< (length age-ann) age-len) ; left-pad
```
alphapapa, line 898: Is it necessary to pad this way?  `format` should be able to handle that.

alphapapa (changes requested):
Thanks for working on this.  Here are a few comments.

activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+                                        (t units)))
+                      (fn age (cdr spec)))))))
+    (fn age activities--age-spec)))
+
+(defun activities--oldest-age (activities)
+  "Return the age in seconds of the oldest activity in ACTIVITIES."
+  (cl-loop for (_name . act) in activities maximize
+	   (cl-loop
+	    for type in '(default last)
+	    for func = (intern (format "activities-activity-%s" type))
```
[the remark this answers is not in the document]
jdtsmith: Forgot about that one, thanks.

activities.el
```diff
@@ -651,6 +652,23 @@ activity's name is NAME."
...
+		    (`(leaf . ,_attrs)
+		     (push (funcall func state) ret))
+		    ((pred atom) state)
+		    (`(,_key . ,(pred atom)) state)
+		    ((pred list)
+		     (if-let ((leaf-pos (cl-position 'leaf state)))
+			 (push (funcall func (cl-subseq state leaf-pos)) ret)
+		       (dolist (s state) (map-leafs s func)))))))
+      (map-leafs state func))
+    (nreverse ret)))
```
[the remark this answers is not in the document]
jdtsmith: You mean branches in the `pcase`, yes?  One of those is necessary to "do nothing" on "uninteresting lists" that may be present.  

BTW, this revealed a bug in your `pcase` formulation (which I borrowed from `bufferize-window-state`): `(pred list)` is always true.  I think `list` is not the predicate you need; I used `proper-list-p`.

activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+
+(defun activities-annotate (max-age oldest-possible)
+  "Return an activity annotation function.
+MAX-AGE is the maximum age of any activity in seconds.
+OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
+  (lambda (name)
+    (when-let ((activity (map-elt activities-activities name)))
+      (let (data (age-len 14))
+	(dolist (type '(last default))
+	  (let* ((func (intern (format "activities-activity-%s" type)))
```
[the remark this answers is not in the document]
jdtsmith: Done.  Should also decide if we want to allow abbreviated ages (or user-customization of the age "words"), in which case padding will need to change.

activities.el
```diff
@@ -801,6 +819,90 @@ activity's name is NAME."
...
+		 (state (funcall func activity))
+		 (time (map-elt (activities-activity-state-etc state) 'time))
+		 (window-state (activities-activity-state-window-state state))
+		 (buffers (window-state-buffers window-state))
+		 (files ( activities--map-window-state-leafs window-state
```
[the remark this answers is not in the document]
jdtsmith: It's actually not extraneous: it prevents emacs-lisp-mode from indenting the subsequent line way over to the right under `window-state`.   It's a reasonably new feature.  But it is ugly, so I hacked it out.  How do you deal with this problem (I notice you often keep the 1st argument on the line with the function call)?

activities.el
jdtsmith: Yes indeed, doh.

jdtsmith:
BTW, are there other pieces of information inside an activity that would be relevant to show that occur to you?

activities.el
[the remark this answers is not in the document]
alphapapa: You're right that it was a mistake: `list` is not a predicate.  `listp` is what I should have used.  It may be that it made no difference in practice.  

I think using `proper-list-p` probably isn't necessary over `listp`; the values in the window state object are well-defined, and misusing `list` was apparently working.

activities.el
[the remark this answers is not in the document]
alphapapa: I don't think customizing that will be necessary.  If it were, there are about 5,000 other places in Emacsland that would be a higher priority.  :)

activities.el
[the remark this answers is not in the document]
alphapapa: Yes, but it's intended for, e.g. quoted plists, not for function calls.  I guess one could use it if one preferred, but it would be unidiomatic and visually confusing, as it would tend to obscure that it's a function call.

I don't like having one-symbol lines, so I avoid them when possible.  It's a balancing act between that and line length.

It's generally fine to fix such places manually, but in my projects I use `makem.sh` with its `lint-indentation` rule, which will give a warning about manually adjusted indentation, so please just indent it as Emacs automatically does and we'll live with it.

alphapapa:
> BTW, are there other pieces of information inside an activity that would be relevant to show that occur to you?

I can't think of any.  For myself, the only one that matters is last-used time, anyway.  Let's keep this as simple as possible.

activities.el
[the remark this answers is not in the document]
alphapapa: Oh, I see how you removed the other clause, so now using `proper-list-p` is likely necessary.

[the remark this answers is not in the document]
jdtsmith: It made no difference I think because all other cases are caught by your prior cases, so `t` would have even sufficed.   Probably best for safety to use `listp`.  I'm using `proper-list-p` because I need to exclude cons cells like `(min-width . 8)` which I no longer match affirmatively.  I could restore that "do-nothing" case or just insist on a "real list"; happy with whatever would make more sense to you down the road.  `cl-position` does not take kindly to non-nil-terminated cons cells.

activities.el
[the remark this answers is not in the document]
jdtsmith: I do agree it's ugly, possibly uglier than one-symbol lines, but I tend to air on the compactness=readability side.  Restored the 1sl for now to be sure not to add linting errors.

jdtsmith:
I could show both default and last ages, but that might be confusing.   

Along that line, do you think having the default and last buf+file cnt is useful or TMI?  Another approach would be to use some indicator (like `*`) that the state is "dirty" in the sense that its display buffer list has changed from the default.   So buffer cnt + file cnt + dirty flag + last mod age.

alphapapa:
> I could show both default and last ages, but that might be confusing.

I don't think that's necessary.  This is just for completing-read, not for "management."

> 
> Along that line, do you think having the default and last buf+file cnt is useful or TMI? Another approach would be to use some indicator (like `*`) that the state is "dirty" in the sense that its display buffer list has changed from the default. So buffer cnt + file cnt + dirty flag + last mod age.

For me, it's TMI.  I select activities by name, not by a number of how many windows or files are visible in it.  I can't think of a situation where I would need to know that.

However, knowing whether an activity has a last-used state could be useful, I think.

jdtsmith:
> > Along that line, do you think having the default and last buf+file cnt is useful or TMI? Another approach would be to use some indicator (like `*`) that the state is "dirty" 
> For me, it's TMI.

I do think I agree.  I like having the last buf/file count, since I use multiple "similar" activities, some richer than others (based on screen size).  But for the default counts, I'm just comparing to see if they changed at all (i.e. computing a "dirty" flag in my head).  For such a flag, what would you recommend?  I was thinking "same number of buffers and same buffer names = not dirty".

alphapapa:
I'm not sure what you mean by "dirty."  For me what would be interesting is whether the activity has a last-used state or just the default.

jdtsmith:
Sorry, I'm not sure I understand your idea.  If you have activity auto-saving on (which is the default), an activity will pretty much always have a "last" state, even if it's only lightly modified from the default (different window points, sizes, etc.).   So my idea of "dirty" is if the last state differs is an interesting way from the default — i.e. different buffers showing.  

I did need to accommodate a missing "last" state, but that's a pretty rare case for me, like when I've just defined a new state.

alphapapa:
> If you have activity auto-saving on (which is the default), an activity will pretty much always have a "last" state, even if it's only lightly modified from the default (different window points, sizes, etc.) ... I did need to accommodate a missing "last" state, but that's a pretty rare case for me, like when I've just defined a new state.

I often use `activities-kill`, which means that the activity won't have a last-used state.

> So my idea of "dirty" is if the last state differs is an interesting way from the default — i.e. different buffers showing.

I feel like that concept might be hard to communicate to the user, but if you want to add an asterisk or something and document it to mean that, I won't object.

jdtsmith:
OK sounds good; lack of asterisk would also mean "no last state".

One oddity I've found: when I `revert` an activity (which I use instead of kill), I'd expect its last state to be removed.  It isn't, and sometimes as a result there are two distinct sets of buffer names for the last and default state, even right after `revert` runs.  This happens if a buffer is a duplicate which has a `uniquify`d name.   I.e. last might have `main.c` while default has `main.c<proj>`.  Have you seen this?  Perhaps revert should just remove last, like kill does.

alphapapa:
> One oddity I've found: when I `revert` an activity (which I use instead of kill), I'd expect its last state to be removed. It isn't, and sometimes as a result there are two distinct sets of buffer names for the last and default state, even right after `revert` runs. This happens if a buffer is a duplicate which has a `uniquify`d name. I.e. last might have `main.c` while default has `main.c<proj>`. Have you seen this? Perhaps revert should just remove last, like kill does.

I can see pros and cons on either side.  Please feel free to open a new issue about that if you like.

jdtsmith:
Thanks.  I've pushed the `*` flag and a README update.  Let me know if you see anything else that needs addressing.

activities.el
```diff
@@ -858,44 +859,44 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
...
+			   window-state
+			   (lambda (leaf)
+			     (bookmark-get-filename
+			      (activities-buffer-bookmark
+			       (map-nested-elt (cdr leaf)
+			 		       '(parameters activities-buffer))))))))
+	      (setf (alist-get type data)
+		    (list (length (delq nil files))
+			  (and time (float-time (time-since time))) buffers)))))
+	(pcase-let* ((`(,last-file-cnt ,last-age ,last-bufs) (map-elt data 'last))
```
alphapapa, line 876: Forgive the nitpicking, but I'd still prefer not to abbreviate these variable names.  I'd suggest, e.g. `num-last-files`, `last-buffers`, etc.

activities.el
```diff
@@ -858,44 +859,44 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
...
+		     (`(,default-file-cnt ,default-age ,default-bufs) (map-elt data 'default))
+		     (age (if last-age (min last-age default-age) default-age))
+		     (buf-cnt (length (or last-bufs default-bufs)))
+		     (file-cnt (or last-file-cnt default-file-cnt))
+		     (dirty (and last-bufs
```
alphapapa, line 881: In Emacs conventions, a flag variable like that should end in `p`, i.e. `dirtyp`.

Also, I prefer to use `when` rather than `and` here, as it clearly distinguishes the condition and the value forms.

activities.el
```diff
@@ -858,44 +859,44 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
...
+				 (or (/= (length last-bufs) (length default-bufs))
+				     (not (seq-set-equal-p last-bufs default-bufs)))))
+		     (ann (format "%s bufs %s files "
+				  (propertize (format "%2d" buf-cnt) 'face 'success)
+				  (propertize (format "%2d" file-cnt) 'face 'warning)))
```
alphapapa, line 886: Again, nitpicking, but I prefer, e.g. `annotation` to `ann`, and `num-buffers` to `buf-cnt`.

alphapapa (changes requested):
Thanks.  Also, could you post a final screenshot showing what it looks like with the asterisk?

activities.el
```diff
@@ -858,44 +859,44 @@ OLDEST-POSSIBLE is the oldest age in the `vc-annotate-color-map'."
...
+		   (buffers (window-state-buffers window-state))
+		   (files (activities--map-window-state-leafs
+			   window-state
+			   (lambda (leaf)
+			     (bookmark-get-filename
+			      (activities-buffer-bookmark
+			       (map-nested-elt (cdr leaf)
+			 		       '(parameters activities-buffer))))))))
+	      (setf (alist-get type data)
+		    (list (length (delq nil files))
```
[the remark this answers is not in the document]
jdtsmith: `window-state-leafs` actually just pushes the results of the function it was provided, which in this case involves `bookmark-get-filename`, which returns nil for non-file buffer bookmarks.   I think it's useful to preserve the nils in case you ever want to correlate the list against other calls.

activities.el
jdtsmith: Not at all!  You'll look at this more than me, so whatever names area easiest on your eyes is preferred.

jdtsmith:
Here's how it's looking (note the final `*` for modified activities):

image

Couple things I think still need a bit of thought: 

1. Active activities get saved automatically, which always updates their `time`.  I guess that could be considered OK, since it shows the activity being continuously saved, but when there are no changes to any of the activity's configuration, it is confusing that the activity age is updated.   I suppose we could replace the age annotation information with `<active>` for all active activities.
2. I had assumed activities were being sorted most recent first, since that's how mine had sorted.  But as the screenshot shows, this isn't the case.  Easy to do by pre-sorting the names based on recency.  Let me know if that should be a separate PR.

alphapapa:
Thanks for the screenshot.

> Couple things I think still need a bit of thought:
> 
>     1. Active activities get saved automatically, which always updates their `time`.  I guess that could be considered OK, since it shows the activity being continuously saved, but when there are no changes to any of the activity's configuration, it is confusing that the activity age is updated.   I suppose we could replace the age annotation information with `<active>` for all active activities.

Well, I do think that active activities should be marked in some way; it would seem like an obviously missing feature otherwise, IMO.

Replacing the age with some other string to indicate its being active would be one possibility.  OTOH the multiple colors already make that column a bit busy-looking.  I'm not sure what the best solution is.

We should probably also keep in mind potential window width in terms of the multiple columns of information.  Have you tested with a half-screen-width window?

>     2. I had assumed activities were being sorted most recent first, since that's how mine had sorted.  But as the screenshot shows, this isn't the case.  Easy to do by pre-sorting the names based on recency.  Let me know if that should be a separate PR.

I think a customizable sorting option should be part of this PR, if you're willing to work on it.  Having the age column but not sorting by it seems weird.  I'd guess that active ones should be sorted first, then the rest by age.

There's also https://github.com/alphapapa/frecency.el, which might be good to use.  But I think that library should be rewritten; I've learned a lot since I wrote it, and its API is a bit awkward, so I wouldn't suggest using it now.  But if the sorting function were customizeable, it could be used later.

jdtsmith:
> I do think that active activities should be marked in some way; it would seem like an obviously missing feature otherwise, IMO.

Yeah, agree.  Perhaps an `<A>` before the buffer count?  Or maybe better, change the codes up:

- `*`=active, modified
- `+`=inactive, modified
- `-`=active, unmodified 
- ` `=inactive, unmodified

> Have you tested with a half-screen-width window?

With the flex-spacer it's responsive to small widths: 

image

If we like it better, could measure the maximum activity width and pile everything up left (but still aligned).  I tend to prefer flush right.  

Will work on a simple sorting scheme.

jdtsmith:
Something else: `window-state-buffers` differ depending on whether that state is loaded and active, or not:

1. Buffers which get renamed by uniquify are updated in the `last` state to their `name<dir>` versions.
2. Buffers visiting symlinked files are named after the symlink (e.g. `.emacs`) in `default`, but the resolved file (e.g. `init.el`) in `last` (or maybe whichever state was loaded).

This leads to false-positive modified states.  Both of these mean I should check for equivalent _files_ (if available) or buffer names (if not).

jdtsmith:
OK I've got something I'm reasonably happy with, with `@` meaning active, `*` meaning modified.  I've also fixed the modification check to compare true filenames and then buffers as a backup, and it fixes the 2 corner cases mentioned above.  Here it is for the absurdly small window width of 45 chars:

image

Obviously still need to sort, and couple of small structural bothers: 

1. When you revert, since last is not immediately updated or removed, the activity still shows up as modified for some time (see #86).  
2. As mentioned above, active activities have their timestamps continuously incremented, so they will _all_ mention the same 5 seconds age (or something small).  It would be better of course to update the timestamp (and data) for a given activity only if something actually changed in the window state.  I confirmed that `cl-tree-equal` works on window-states.

jdtsmith:
I got sort working, but I forgot that vertico substitutes it's own "most recent used then shortest candidate" sort function.  You can disable it by let-binding `vertico-sort-function nil` but that's a bit odd to throw in.  Thoughts?

jdtsmith:
Have you ever seen the persist file disappear?  It's happened to me twice in the past few days while hacking on this.  I note that persist removes the file if the value is ever the default value on save (`nil`).  I don't see any reason it would be getting set to nil, but perhaps edebug is involved.   Auto-save is on.

alphapapa:
> Have you ever seen the persist file disappear? It's happened to me twice in the past few days while hacking on this. I note that persist removes the file if the value is ever the default value on save (`nil`). I don't see any reason it would be getting set to nil, but perhaps edebug is involved. Auto-save is on.

It has happened to me with another package, but only when I used my `unpackaged/reload-package` command to reload the package's symbols after it was already loaded.  Since I stopped doing that, it hasn't happened again.  I'm guessing there's some kind of unexpected state and actions that can happen when developing if it involves reloading the file in which the variable is defined, or if `C-M-x` is used on the variable's definition.  So AFAICT it shouldn't ever affect users, but developers could find it quite inconvenient.  It would probably be worth filing an issue about and trying to nail it down and possibly fix it.

jdtsmith:
OK, see #87.  Give the sort and active marker a look and let me know what you think about preventing the time-tick from updating unless "things have changed".

jdtsmith:
Anything further needed here?  If you like I can look into updating activity timestamp on save only if buffers changed.

alphapapa:
I have a lot going on right now, so it may be a while before I can review this further.  I appreciate your patience.

jdtsmith:
No rush on review, but I wanted to mention that I factored out the buffer/file comparison code from the annotation function, and  updated `activities-save` using it to retain the timestamp if the list of buffers/files has not changed.  This works _much_ better with the idle-time `save-all`, no longer spuriously updating the age of active but "idle" activities.   Working quite well so far.

jdtsmith:
Gentle ping on this one.  Been in active use for quite some time without any issues on my end.

alphapapa:
> Gentle ping on this one. Been in active use for quite some time without any issues on my end.

Thanks.  My to-do list is quite long and my time has been limited.  I guess this can be merged now...

activities.el
```diff
@@ -801,6 +851,116 @@ activity's name is NAME."
...
+    (?w "week"   "weeks"   ,(* 60 60 24 7))
+    (?d "day"    "days"    ,(* 60 60 24))
+    (?h "hour"   "hours"   ,(* 60 60))
+    (?m "minute" "minutes" 60)
+    (?s "second" "seconds" 1))
+  "Age specification.  See `magit--age-spec', which this duplicates.")
+
+(defun activities--age (age &optional abbreviate)
+  "Summarize AGE and possibly ABBREVIATE.
+Adapted from `magit--age'."
```
alphapapa, line 866: I'm not sure if this can be adapted here, because Activities is in GNU ELPA and is copyrighted by the FSF, while Magit is not copyrighted by the FSF.  @tarsius Could you please advise here?

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
...
+		     (if-let ((leaf-pos (cl-position 'leaf state)))
+			 (push (funcall func (cl-subseq state leaf-pos)) ret)
+		       (dolist (s state) (map-leafs s func)))))))
+      (map-leafs state func))
+    (nreverse ret)))
+
+(defun activities--buffers-and-files (state)
+  "Return a list of buffers and files from STATE.
+STATE is a window-state.  The returned list contains elements of
+form (BUFFER . FILE) assocaited with the activity."
```
alphapapa, line 535: Nitpick: Spelling typo.  :)

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
 
 ;;;; Functions
 
+(defun activities--map-window-state-leafs (state func)
+  "Call FUNC on all the leaf nodes in window-state STATE.
+The resulting values are returned as a list."
+  (let (ret)
```
alphapapa, line 520: Nitpick: Could we call this variable `values` instead of `ret`?

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
 
 ;;;; Functions
 
+(defun activities--map-window-state-leafs (state func)
+  "Call FUNC on all the leaf nodes in window-state STATE.
```
alphapapa, line 518: I prefer for the first line to say what the function returns.  How about something like, "Return a list of values of having applied FUNC to leaf nodes in STATE."?  Maybe the function could be called `activities--mapcar-window-state-leafs` to additionally imply that it returns a list like `mapcar`.  What do you think?

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
...
+    (nreverse ret)))
+
+(defun activities--buffers-and-files (state)
+  "Return a list of buffers and files from STATE.
+STATE is a window-state.  The returned list contains elements of
+form (BUFFER . FILE) assocaited with the activity."
+  (activities--map-window-state-leafs
+   (activities-activity-state-window-state state)
+   (lambda (leaf)
+     (let ((rec (map-nested-elt (cdr leaf) '(parameters activities-buffer))))
```
alphapapa, line 539: Could we use a more descriptive variable name than `rec` here?

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
...
+       (cons (activities-buffer-name rec) (activities-buffer-filename rec))))))
+
+(defun activities--buffers-and-files-differ-p (bf1 bf2)
+  "Return t if BF1 and BF2 are not the same set of files or buffers."
```
alphapapa, line 543: This should probably say "Return non-nil if...", because it shouldn't matter whether it's `t` or another non-nil value.

Also, it would be good to explain what BF1 and BF2 are (checkdoc will complain otherwise, at least).

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
...
+  (cl-labels ((file-or-buffer (cell)
+		"Given (buffer . file), return the true filename or (if none) buffer."
```
alphapapa, line 545: I like having the docstring in the label function, but it should probably follow the convention if it's to be there.

activities.el
```diff
@@ -503,6 +514,39 @@ To be called from `kill-emacs-hook'."
...
+		(if (cdr cell) (file-truename (cdr cell)) (car cell))))
```
alphapapa, line 546: Should we use `abbreviate-file-name` as well?

activities.el
```diff
@@ -801,6 +851,116 @@ activity's name is NAME."
...
 
+(defvar activities--age-spec
+  `((?Y "year"   "years"   ,(round (* 60 60 24 365.2425)))
+    (?M "month"  "months"  ,(round (* 60 60 24 30.436875)))
+    (?w "week"   "weeks"   ,(* 60 60 24 7))
+    (?d "day"    "days"    ,(* 60 60 24))
+    (?h "hour"   "hours"   ,(* 60 60))
+    (?m "minute" "minutes" 60)
+    (?s "second" "seconds" 1))
+  "Age specification.  See `magit--age-spec', which this duplicates.")
```
alphapapa, line 862: See copyright question below.

activities.el
```diff
@@ -801,6 +851,116 @@ activity's name is NAME."
...
+							    (activities--age age)))
+				      'face `(:foreground ,age-color
+							  :background ,vc-annotate-background)))
+		     (dirty-annotation (if dirtyp (propertize "*" 'face 'bold) " ")))
+	  (concat (propertize " " 'display
+			      `(space :align-to (- right ,(+ 1 (length annotation)
+							     (length age-annotation)))))
+		  annotation age-annotation dirty-annotation))))))
+
+(defun activities-sort-by-active-age (names)
```
alphapapa, line 935: If this function is only used in one place, I'd like to avoid defining it as a top-level symbol.

activities.el
```diff
@@ -801,6 +851,116 @@ activity's name is NAME."
...
+		 (time-a (map-elt (activities-activity-state-etc state-a) 'time))
+		 (activep-a (activities-activity-active-p activity-a))
+		 (activity-b (map-elt activities-activities b))
+		 (state-b (or (activities-activity-last activity-b)
+			      (activities-activity-default activity-b)))
+		 (time-b (map-elt (activities-activity-state-etc state-b) 'time))
+		 (activep-b (activities-activity-active-p activity-b)))
+	    (or (and activep-a (not activep-b)) (time-less-p time-b time-a))))))
+
+(defun activities--completion-table (activities)
```
alphapapa, line 951: Do you foresee this function being called elsewhere than in `activities-completing-read`?  If not, I think I'd prefer for it to be defined inside that function, to avoid the top-level symbol.

activities.el
```diff
@@ -801,6 +851,116 @@ activity's name is NAME."
...
+  "Return the age in seconds of the oldest activity in ACTIVITIES."
+  (cl-loop for (_name . act) in activities maximize
+	   (cl-loop
+	    for type in '(default last)
+ 	    for state = (cl-struct-slot-value 'activities-activity type act)
+	    if state
+	    for etc = (activities-activity-state-etc state)
+	    maximize (float-time (time-since (map-elt etc 'time))))))
+
+(defun activities--annotate (max-age oldest-possible)
```
alphapapa, line 888: Forgive my laziness, since my time is limited, but could this be done without defining this top-level symbol?  I really like to try to avoid defining new ones.

activities.el
```diff
@@ -66,6 +66,7 @@
 (require 'map)
 (require 'persist)
 (require 'subr-x)
+(require 'vc-annotate)
```
alphapapa, line 69: OT1H this seems like an elegant reuse of code; OTOH many users would never load this library otherwise (myself being one of them).  Is there another way we could do the equivalent, while still reusing some face/color definitions?

alphapapa (changes requested):
Well, apologies for the long delay, and then asking for more changes.  Thanks for your work on this.  It will be merged in the end.  :)

activities.el
tarsius: That's correct. However, I am the only one who ever touched this function and the associated variable, and would be happy to contribute this to Emacs. I would ask @jdtsmith to do the work of getting it integrated (or, if that's what the maintainers request, replaced by a better implementation).

alphapapa: @tarsius Thanks.  With your being the author, if you consent, then, can it just be "contributed to Emacs" via this package, since ELPA packages are "part of GNU Emacs"?

tarsius: Well yes, but let's wait and see if @jdtsmith would be willing to get it into Emacs first. Doesn't have to be this implementation, but something like this would certainly be useful in more than these two packages.

As for the used algorithm, one might want to consider using another threshold than 1. I don't really mind that 2.0 and 2.49 years are both "2 years", but when it comes to 0.51 and 1.49 both being "1 year", I feel a bit differently. Maybe this should be customizable, maybe even per time unit.

If I remember correctly, I did shop around for other implementations before writing mine, but the once I found all addressed this by using more than one time unit, which doesn't work in our limited space constraint.

jdtsmith: Happy to help out.  What did you envision?  I also use some color and date-related functionality from `vc-annotate`, perhaps it could go there (though I rarely use `vc` anymore, because magit ;).  

A little more granularity would be nice; maybe a unicode ½ could enter.

jdtsmith:
> > Gentle ping on this one. Been in active use for quite some time without any issues on my end.
> 
> Thanks. My to-do list is quite long and my time has been limited. I guess this can be merged now...

Happy to move at your pace, no problem.  I tend to forget such things myself and so appreciate a ping here and there.

activities.el
jdtsmith: Done.

activities.el
jdtsmith: There is a customization option `activities-sort-function` for which this function is the default value.   This allows the user to specifying another sort.

activities.el
jdtsmith: These files (or buffers) are used exclusively to determine is an activity has changed, so abbreviating the files wouldn't add anything I don't think.  They don't get displayed.

activities.el
jdtsmith: I could shove this into `activities-completing-read` just as I did the completion table, but it's a pretty meaty function.  What do you think?

activities.el
jdtsmith: I've used them before in other packages.  In fact I know about these colors because a better color-age dynamic range mapping was the first contribution I made to emacs (in a prior millennium).  We could rip the colors out of there and duplicate, but that colormap has been around long enough that I wonder if anyone has customized it.  It is used for blame functionality, so quite similar in spirit.  

It's trivial to select a precise color based on fractional age, using `(color-hsl-to-rgb)`, perhaps also giving users the ability to pick a lightness/saturation value in a defcustom.   Thoughts?

E.g. 
image

alphapapa: I'd suggest perhaps using this function I wrote for `ement`, which does something very similar but a bit more nuanced: https://github.com/alphapapa/ement.el/blob/9f6cbb8b406160c950d8932d66f3629aba091ec8/ement-room-list.el#L849-L890

jdtsmith: I eliminated vc-annotate, and went with a simple blend from a specified old to new color, and a blend fraction with the default foreground.

activities.el
jdtsmith: I managed to squeeze the annotation function in there.

activities.el
jdtsmith: I think the right approach is to get something like this into time-date.el, copy it here temporarily, and once it shows up in `compat`, remove it.  I'll send in a patch.

jdtsmith: One issue is this similar existing function:

```elisp
(defvar seconds-to-string
  (list (list 1 "ms" 0.001)
        (list 100 "s" 1)
        (list (* 60 100) "m" 60.0)
        (list (* 3600 30) "h" 3600.0)
        (list (* 3600 24 400) "d" (* 3600.0 24.0))
        (list nil "y" (* 365.25 24 3600)))
  "Formatting used by the function `seconds-to-string'.")
;;;###autoload
(defun seconds-to-string (delay)
  ;; FIXME: There's a similar (tho fancier) function in mastodon.el!
  "Convert the time interval in seconds to a short string."
  (cond ((> 0 delay) (concat "-" (seconds-to-string (- delay))))
        ((= 0 delay) "0s")
        (t (let ((sts seconds-to-string) here)
             (while (and (car (setq here (pop sts)))
                         (<= (car here) delay)))
             (concat (format "%.2f" (/ delay (car (cddr here)))) (cadr here))))))
```

activities.el
tarsius: > A little more granularity would be nice; maybe a unicode ½ could enter.

My 2 cents: Sure ½ years is nice, but 12½ years is unnecessary "precision". I like this progression of half year steps better 6 months, 12 months, 18 months, 2 years, 3 years (for 2.5), 3 years (for 3), and so on. This is actually more precise (in the range were it matters) than the ½ approach: 4/6/8 months vs. ½/½/½ years. And that's what I don't like about the half-step approach, it suggests precision that just isn't there.

jdtsmith: The progression would go 1 hour, 1½ hours, 1 day, 1½ days, 1 week, 1½ weeks, 1 month, 1½ months, 6 months, 1 year, 1½ years.  I can see an argument for dropping the ½ and reverting to integer rounding after a certain base value, perhaps 3½?  What about 7½ months vs. 7 or 8 months?  I could go either way on that one.  Most relevant for years (I briefly toyed with adding "decades" but... no).  I agree that 29½ years is fairly silly.   The current `seconds-to-string` uses 1/100ths across the board btw.

activities.el
jdtsmith: See bug#71573.

activities.el
tarsius: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71572

tarsius: I think I am going to stick what I have now.

> The current seconds-to-string uses 1/100ths across the board btw.

Had I known that function and if it used 1/10, I probably would have used it.

A code comment mentions that there is a fancier implementation to be found in `mastodon.el`. I wasn't able to locate that, could you?

jdtsmith: No I didn't.  Take a look at the patch I submitted (should also be in your inbox).  Happy for further  input.

activities.el
alphapapa: FWIW, I replied to that bug report linking to this code in my `ts` library: https://github.com/alphapapa/ts.el/blob/master/ts.el#L440-L491  Although it works a bit differently, I've used it extensively in some of my other packages, and profiling shows that it performs very well in terms of runtime and GC.  If any of it seems useful, I'll be glad to contribute it upstream as well (I'd be glad to contribute all of `ts`, in fact, but I haven't exactly proposed that yet).

jdtsmith: Thanks.  My guess is the attention to cutting v30 will mean this sits for a while.  I tried to adapt to the style of date-time but am not committed to any particular version. I would like to include weeks and months; to me that's the key elaboration of `magit--age`
(plus the modest resolution).  

How about we give it a week and if no movement, put in a version ourselves, with the aim to remove it once the patch is included in a release (+compat)?

activities.el
tarsius: There obviously are different approaches to this task. For the reasons given, I don't like the half-step approach much, and looking at the implementation did not change that. That shouldn't keep you from iterating on it. There's room for different approaches. If something lands in Emacs, it would probably be a good idea, if it were done in a way that allowed for switching in an alternative implementation. (I.e., using a `...-function` variable, though in the absence of that, an `:override` advice would of course also work.)

jdtsmith: A good design goal IMO is "something tarsius would be happy to sub in for magit--age at some point", so feel free to express your ideas [there](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71573).  So far no response.  `HALF` is optional btw, default off.

activities.el
jdtsmith: Some uptake on the patch adding a new `readable` option to `seconds-to-string`, but Eli just confirmed it wouldn't land until Emacs 31, so I propose we include a copy here along the same lines, with a note to replace it eventually.  Pushed changes along these lines.

activities.el
```diff
@@ -332,6 +333,26 @@ Kills buffers that have only been shown in that activity's
...
 
+(defcustom activities-sort-function #'activities-sort-by-active-age
+  "A function to use to sort by when prompting for activities.
+If nil, no sorting will be applied.  The function should take two
+arguments, both activity names (strings).  It should return
+non-nil if the first activity should sort before the second.  By
+default, a function is used which sorts active activities first,
+and then by age since modification."
+  :type '(choice (const :tag "No sorting" nil)
+		 (function :tag "A specific function")))
```
alphapapa, line 344: I think this could be a bit more idiomatic by describing the option in terms of the result rather than the predicate used to achieve it.  For example, naming the option `activities-sort-by`.  Then the type spec should include the default predicate, and the tag would be, e.g. "Age", so the user could choose to sort by "Age" or "custom predicate."

The docstring needn't describe the predicate beyond its arguments, I think, since that's just how a sort predicate works.

activities.el
```diff
@@ -332,6 +333,26 @@ Kills buffers that have only been shown in that activity's
...
+
+(defcustom activities-annotation-colors '("blue" "red" 0.65)
+  "Colors to use for annotating activity age.
+A list (OLD-COLOR NEW-COLOR BLEND-FRAC).  These are used during
+activity selection, with the activity color chosen between
+OLD-COLOR and NEW-COLOR, based on the activity's age.  This color
+is blended into the default foreground with a fraction
+BLEND-FRAC, and used to display the activity age."
+  :type '(list (color :tag "Old Color") (color :tag "New Color")
+	       (float :tag "Blend Fraction")))
```
alphapapa, line 354: It might be more idiomatic to call the float the "alpha" or "opacity".  Also, it would be good to make the docstring a bit more concise.

activities.el
```diff
@@ -503,6 +524,45 @@ To be called from `kill-emacs-hook'."
...
+(defun activities--buffers-and-files-differ-p (bf1 bf2)
+  "Return non-nil if BF1 and BF2 are not the same set of files or buffers.
+Each of BF1 and BF2 is a list of buffer and files, as returned
+from `activities--buffers-and-files'."
+  (cl-labels ((file-or-buffer (cell)
+		"Given a CELL, return the true filename or buffer.
+The CELL is a (BUFFER . FILE) cons.  If the file is nil, BUFFER is returned."
+		(if (cdr cell) (file-truename (cdr cell)) (car cell))))
+    (not (seq-set-equal-p (mapcar #'file-or-buffer bf1)
+			  (mapcar #'file-or-buffer bf2)))))
```
alphapapa, line 564: Very nitpicky, but I'd prefer to name these arguments, e.g. `bf-a` and `bf-b`, because they aren't sequential in nature.  (Maybe just an idiosyncrasy of how my mind works.)

activities.el
```diff
@@ -801,6 +867,63 @@ activity's name is NAME."
...
+    (?d "day"    "days"    ,(* 60 60 24))
+    (?h "hour"   "hours"   ,(* 60 60))
+    (?m "min"    "mins"    60)
+    (?s "sec"    "secs"    1))
+  "Age specification.  See `magit--age-spec', which this duplicates.")
+
+(defun activities--age (age &optional abbrev)
+  "Summarize AGE.
+Abbreviate the units if ABBREV is non-nil.  Based on
+`magit--age'."
```
alphapapa, line 883: It's important to mention the provenance of the code, but that can go in a comment rather than the docstring.

activities.el
```diff
@@ -801,6 +867,63 @@ activity's name is NAME."
...
+  ;; TODO: replace this if seconds-to-string adds READABLE support
```
alphapapa, line 884: Does this TODO need to be updated?  Also, a link to the relevant Emacs bug should be included.

activities.el
```diff
@@ -801,6 +867,63 @@ activity's name is NAME."
...
+	  (let* ((activity-a (map-elt activities-activities a))
+		 (state-a (or (activities-activity-last activity-a)
+			      (activities-activity-default activity-a)))
+		 (time-a (map-elt (activities-activity-state-etc state-a) 'time))
+		 (activep-a (activities-activity-active-p activity-a))
+		 (activity-b (map-elt activities-activities b))
+		 (state-b (or (activities-activity-last activity-b)
+			      (activities-activity-default activity-b)))
+		 (time-b (map-elt (activities-activity-state-etc state-b) 'time))
+		 (activep-b (activities-activity-active-p activity-b)))
```
alphapapa, line 924: Would you be interested in using `pcase-let*` to do that binding?  :)  If not, that's okay; just an idea.

activities.el
```diff
@@ -801,6 +867,63 @@ activity's name is NAME."
...
+	    (or (and activep-a (not activep-b)) (time-less-p time-b time-a))))))
```
alphapapa, line 925: This would be a bit easier to read if the two OR'ed forms were on separate lines.

activities.el
```diff
@@ -809,9 +932,60 @@ activity's name is NAME."
...
 PROMPT is passed to `completing-read' by way of `format-prompt',
 which see, with DEFAULT."
-  (let* ((prompt (format-prompt prompt default))
-         (names (activities-names activities))
-         (name (completing-read prompt names nil t nil 'activities-completing-read-history default)))
+  (pcase-let*
+      ((names (activities-names activities))
+       (max-age (activities--oldest-age activities))
+       (`(,old-col ,new-col ,blend-frac) activities-annotation-colors)
+       (annotation-function
```
alphapapa, line 939: FWIW, I'll probably move the annotation function into a `cl-labels` form; just a preference of mine.

README.org
```diff
@@ -135,6 +135,15 @@ Key bindings are, as always, ultimately up to the user.  However, in [[Configura
 
 When option ~activities-bookmark-store~ is enabled, an Emacs bookmark is stored when a new activity is made.  This allows the command ~bookmark-jump~ (~C-x r b~) to be used to resume an activity (helping to universalize the bookmark system).
 
+** Annotation and Sort
```
alphapapa, line 138: This section should probably be titled "Completion".

README.org
```diff
@@ -135,6 +135,15 @@ Key bindings are, as always, ultimately up to the user.  However, in [[Configura
...
+
+When selecting an activity in the mini-buffer, annotations are sorted active first, then by most recent modification age.  If you are using an interface that supports /annotations/, additional information is shown alongside the activity name:
+
+- an `@` symbol, for active activities,
+- the number of buffers and files the activity contains,
+- a color-coded age since the activity was last updated, and
+- a final ~*~ if the activity's list of buffers and files has been modified from its default state.
```
alphapapa, line 145: I think the sorting order will be obvious to the user, so probably the only thing that needs to be documented is what the non-obvious symbols mean.

alphapapa (changes requested):
Forgive the additional round of review and comments.  If you don't have enough "steam" to make some of the stylistic changes, I can make them myself before merging.

jdtsmith:
> Forgive the additional round of review and comments. If you don't have enough "steam" to make some of the stylistic changes, I can make them myself before merging.

No worries at all.  In my view, asking an author to incorporate significant changes, and (far more importantly) _support them for years or decades to come_ calls for real patience.  As long as good faith progress is being made towards merge, I'm happy to iterate.

alphapapa:
> > Forgive the additional round of review and comments. If you don't have enough "steam" to make some of the stylistic changes, I can make them myself before merging.
> 
> No worries at all. In my view, asking an author to incorporate significant changes, and (far more importantly) _support them for years or decades to come_ calls for real patience. As long as good faith progress is being made towards merge, I'm happy to iterate.

I appreciate that.  Sometimes I wonder if I should handle some nitpicks myself before merging; I try to consider whether mentioning it will serve any educational purpose, but then I don't always know how familiar the contributor already is, and whether they just have other preferences.

activities.el
jdtsmith: Not sure if you are following that bug (#[71572](https://yhetil.org/emacs-bugs/D91AFAEE-C909-4238-AFF2-226727D4B0F6@gmail.com/T/#t)), but some expressed concern about including a unicode char.  I don't personally think that's an issue in this day and age, but wanted to check.

alphapapa: I've been following it but not very closely.  I didn't realize how much of a bikeshed it was going to become (not a hostile one--I'm just surprised at how long it's gone on).  :)  Maybe that's why there are several implementations floating around, because everyone has their own preferences.

I don't have a problem with using Unicode here, as long as the character is readable, and not only present in very recent versions of common fonts.

activities.el
jdtsmith: Sure see how that looks.  I went for a `cl-labels` an `pcase-lambda` to try to make it a bit more readable.  Eye of the beholder of course (and many stalwart elispers think pcase is de-facto unreadable).

activities.el
jdtsmith: Honestly it does feel we are packing a _lot_ into this function body.  I'd certainly reach for an internal function to keep it clearer, but again what matters is what's clear to _you_.  I converted them both to named local functions with `cl-labels`.  I didn't think I'd like it, but I do like that it's explicit that they are functions now in the call to `completing-read`, and have docstrings for clarity.

jdtsmith:
> > > Forgive the additional round of review and comments. If you don't have enough "steam" to make some of the stylistic changes, I can make them myself before merging.
> > 
> > 
> > No worries at all. In my view, asking an author to incorporate significant changes, and (far more importantly) _support them for years or decades to come_ calls for real patience. As long as good faith progress is being made towards merge, I'm happy to iterate.
> 
> I appreciate that. Sometimes I wonder if I should handle some nitpicks myself before merging; 

I'd say feel free to push any changes yourself if they are glaring and not worth describing.  But I appreciate the collaborative feel your approach creates.  I do a lot of work on other people's packages I admire, and some very capable authors have trouble with that, and can't resist NIH syndrome.  So I'd say your approach is very valuable, especially to any less experienced contributors who come along.

> I try to consider whether mentioning it will serve any educational purpose, but then I don't always know how familiar the contributor already is, and whether they just have other preferences.

I do have some different preferences but I always enjoy learning how others approach a problem.  And I don't mind adapting to the local accent (Canadian ? ;).  So some of both here.

As an example I mentioned above, I don't often reach for `cl-labels` when composing `lambda`'s would do — i.e. when you are just going to pass the function on as an argument, once.  But I found to my surprise I like quite the readability and mimetics (of `some-func` vs. `#'my-local-func`), so may use that more.

Other than deciding whether we are OK with ½ in the ages I think this is ready from my end.  Feel free to season to taste.

jdtsmith:
Actually one more thing: I don't like that `max-age` maximizes across both `last` and `default.`  It should use `last` _or_ `default.`

activities.el
jdtsmith: I fully expected the bike-shedding ;).  We'll see how it converges; it appears it's a problem that's been solved many times.  As long as something reasonably flexible appears in Emacs 31 it probably doesn't matter much.  If it goes in without the ½, which seems likely, we can always give precision = 0.5 and modify `x.5` to read `x½`.

alphapapa:
> I do have some different preferences but I always enjoy learning how others approach a problem. And I don't mind adapting to the local accent (Canadian ? ;). So some of both here.

Do I seem Canadian?  I'm American, born and raised.  :)

`C-h f pcase RET` shows me:
```
-- (cl-struct TYPE &rest FIELDS)

Pcase patterns that match cl-struct EXPVAL of type TYPE.
Elements of FIELDS can be of the form (NAME PAT) in which case the
contents of field NAME is matched against PAT, or they can be of
the form NAME which is a shorthand for (NAME NAME).
```
But that might only show up after loading the relevant library that defines that `pcase` pattern, IIRC.

Also, FYI:
```
-- (cl-type TYPE)

Pcase pattern that matches objects of TYPE.
TYPE is a type descriptor as accepted by ‘cl-typep’, which see.
```

jdtsmith:
> `C-h f pcase RET` shows me:
> 
> ```
> -- (cl-struct TYPE &rest FIELDS)
> 
> Pcase patterns that match cl-struct EXPVAL of type TYPE.
> Elements of FIELDS can be of the form (NAME PAT) in which case the
> contents of field NAME is matched against PAT, or they can be of
> the form NAME which is a shorthand for (NAME NAME).
> ```

Ahh, right you are.  What's strange is there's nothing in `info`, my usual go-to for `pcase.`  In general the info and docstring are similar enough that it's strange what is omitted.  For example, both include `cl-type`, but only the docstring mentions `seq`, `radix-tree-leaf` and `cl-struct`.

alphapapa:
> Ahh, right you are. What's strange is there's nothing in `info`, my usual go-to for `pcase.` In general the info and docstring are similar enough that it's strange what is omitted. For example, both include `cl-type`, but only the docstring mentions `seq`, `radix-tree-leaf` and `cl-struct`.

Probably an oversight, then, and probably worth a bug report.

activities.el
```diff
@@ -332,6 +333,26 @@ Kills buffers that have only been shown in that activity's
...
+(defcustom activities-sort-by #'activities-sort-by-active-age
+  "How to sort activities during selection.
+Function used to sort by when prompting for activities.  By
+default, a function is used which sorts active activities first,
+and then by age since modification.  If nil, no sorting will be
+applied.  A custom predicate function may also be set.  It should
+take two arguments, both activity names (strings), and return
+non-nil if the first activity should sort before the second."
+  :type `(choice (const :tag "No sorting" nil)
+		 (const :tag "Active state and age" ,#'activities-sort-by-active-age)
```
alphapapa, line 345: Rather than `const`, this should probably be `function-item`.  (The details of the customization system are covered in the Elisp manual; there's a lot to it.)

activities.el
```diff
@@ -809,11 +929,65 @@ activity's name is NAME."
...
+		 (concat (propertize " " 'display
+				     `(space :align-to (- right ,(+ 1 (length annotation)
+								    (length age-annotation)))))
+			 annotation age-annotation dirty-annotation)))))
+	 (activity-table (str pred action)
+	   "Complete activities from STR, using completion PRED and ACTION."
+	   (if (eq action 'metadata)
+	       `(metadata (annotation-function . ,#'activity-annotation-function)
+			  ,@(when activities-sort-by
+			      `(,(cons 'display-sort-function activities-sort-by))))
```
alphapapa, line 985: Is it necessary to support sorting being optional?  It seems like disabling sorting would be of no benefit.

alphapapa (changes requested):
Thanks, I think this will be ready to merge after this.

activities.el
jdtsmith: I guess no-sorting is actually "sorting by the order I made these activities", right?  Could be useful to someone.

activities.el
jdtsmith: OK done.  I dislike how `function-item` and others mention the function name in the custom interface, instead of the `:tag` so that's why I used `const`.  But adding `:doc` is a decent workaround.

alphapapa: Hm, that's a good point.  I'll leave this decision up to you, then.  :)

activities.el
alphapapa: I don't think the order of the `activities-activities` list is defined, so "no sorting" would be equivalent to "unknown order," which doesn't seem useful to me.  Unless there's a potential performance issue that would make it necessary to disable sorting, I think I'd rather go for the simpler code that expects the sort-by function to always be set (it could be set to `always` or `ignore` to effectively disable sorting, if the user really wanted).

jdtsmith: Fair enough, the user can always slot `#'ignore` in if they want no change.  I also found a last minute slight flaw in the sort logic and fixed that.

activities.el
jdtsmith: I think it's fine with `:doc`, and certainly more semantically meaningful.

alphapapa:
Thanks for your considerable patience in this process.  One final thing: would you be willing to write a changelog entry for this branch's changes?

jdtsmith:
> Thanks for your considerable patience in this process. One final thing: would you be willing to write a changelog entry for this branch's changes?

No problem, done.

jdtsmith:
Anything more to do on this PR, @alphapapa?

alphapapa:
> Anything more to do on this PR, @alphapapa?

Thanks for your patience.

By the way, forgive me for asking about this again, but I can't find where I asked you specifically before, and this contribution is larger, and I'm trying to be extra diligent about this when accepting contributions: You have completed the FSF copyright assignment for Emacs, right?  I see in emacs.git some contributions from you, but it's not clear whether they are covered by an assignment (e.g. the recent ones under jdtsmith@gmail.com say they are exempt).  

To be certain, I'm required by the Emacs maintainers to ask them for confirmation for a given name and email address.  So is the right name and email address to ask them about `"JD Smith <jdtsmith@gmail.com>"`?  (Again, apologies if this is redundant, but I'm trying to be very careful.)

jdtsmith:
I appreciate your diligence (which no doubt saves headaches down the road).  Yes, I completed this 23 yrs ago ;).  I recently updated my email with the FSF, so that should work for confirmation.

alphapapa:
Thanks!  Email sent.  Should hear back soon, then I can merge this PR.

alphapapa:
Thanks for your patience.  Eli confirmed your CA (a while ago, I just didn't get around to handling the message until now).

jdtsmith:
Making my list and checking it twice... I see this one is still missing a checkmark ;).

alphapapa:
@jdtsmith Thanks, you've been extremely patient.  I'll merge this soon... (even writing a note on paper to remind me)

alphapapa:
@jdtsmith I rebased on master in preparation for merging.  

While doing some final checks, I noticed one final thing to ask about: While your FSF CA was done as `jdtsmith@gmail.com`, these git commits have the address `93749+jdtsmith@users.noreply.github.com`.  Is that intentional?  It would be more convenient for copyright attribution if the same address were used for authoring the commits as used when filing copyright assignment.  Otherwise, if someone goes auditing in the future, one would have to dig up these conversations and try to find the link in the chain between that "noreply.github.com" address and your real address which is on file.

jdtsmith:
I have the "keep email address private" option selected, which means github refuses pushes with your actual email as author:

```
remote: error: GH007: Your push would publish a private email address.        
remote: You can make your email public or disable this protection by visiting:        
remote: https://github.com/settings/emails        
```


But since it's been mentioned here I temporarily changed that and force-pushed with updated author email.  It was a lot of commits!  Luckily magit with rebase-interactive and a quick keyboard macro made quick work of it.

jdtsmith:
BTW, RE our discussions above of richer abbreviated dates, my related patch was [merged to master](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71572) recently, and will soon be added to `compat` (which I notice you do not use?).  For the future.

alphapapa:
> But since it's been mentioned here I temporarily changed that and force-pushed with updated author email. It was a lot of commits! Luckily magit with rebase-interactive and a quick keyboard macro made quick work of it.

Thanks.

So, after about 8 months, this is finally merged.  Thanks for your contribution and, of course, your patience.

> BTW, RE our discussions above of richer abbreviated dates, my related patch was [merged to master](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71572) recently, and will soon be added to compat (which I notice you do not use?). For the future.

I use `compat` in my packages where it's needed.  There hasn't been a need in this one yet.  If it helps, or allows removing code, by all means, let's use it.

PR #98, opened 2024-06-18
windows-set: avoid frame race condition

jdtsmith:
In discussion #78 I suggested a loop over activity names, calling `activities-resume`.  This doesn't work reliably, because of the `HACK` in `activities--windows-set`, which uses an "immediate" timer.  When rapidly resuming activities in a loop, by the time the timer is called, the frame will have changed, and the activities' contents are bungled.

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -250,8 +250,7 @@ discarded, such a bookmark could become stale."
         (cons 'window-side 'writable)
         (cons 'window-slot 'writable))
   "Additional window parameters to persist.
-See Info node `(elisp)Window Parameters'.  See also option
-`activities-set-window-persistent-parameters'."
+See Info node `(elisp)Window Parameters'."
   :type '(alist :key-type (symbol :tag "Window parameter")
                 :value-type (choice (const :tag "Not saved" nil)
                                     (const :tag "Saved" writable))))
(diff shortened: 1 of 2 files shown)


activities.el
```diff
@@ -648,8 +648,9 @@ activity's name is NAME."
     ;; HACK: Since `bookmark--jump-via' insists on calling a buffer-display
     ;; function after handling the bookmark, we use an immediate timer to
     ;; set the window configuration.
-    (run-at-time nil nil (lambda ()
-                           (window-state-put state (frame-root-window) 'safe)))))
+    (let ((frame (selected-frame)))
```
alphapapa, line 651: Thanks, this looks like the right fix.  What if we bind that variable in the `pcase-let*` form above?

alphapapa (changes requested):

activities.el
jdtsmith: Done.  In fact just a simple `let` will do.

jdtsmith:
`window-persistent-parameters` will no longer be let-bound by the time the nil timer runs, so it needs to be set inside the timer callback; implemented that too.

jdtsmith:
Anything else need a look here?

alphapapa:
> Anything else need a look here?

Looking at this code again, now that you've apparently fixed an outstanding bug with the binding of `window-persistent-parameters`, I'm guessing that we should remove `(setf window-persistent-parameters (copy-sequence activities-window-persistent-parameters))` from the top of the function; what do you think?

Also, forgive the nitpicking, but I'd suggest simplifying the `frame` binding one step further by making the lambda take the frame as its argument, and then passing `(selected-frame)` as the argument to the timer.

jdtsmith:
> Looking at this code again, now that you've apparently fixed an outstanding bug with the binding of `window-persistent-parameters`, I'm guessing that we should remove `(setf window-persistent-parameters (copy-sequence activities-window-persistent-parameters))` from the top of the function; what do you think?

I actually don't know what function that copying served vs. the let-binding already happening in `activities--window-state`; maybe it was a workaround to this escaping-the-let-bind bug.  It effectively overrode the user's `persistent-parameters`, which seems non-ideal, if other packages or features used it.  Deleting the copy doesn't seem to cause problems, but then again I don't save any exotic state, so maybe that needs testing.

I also note that `activities-set-window-persistent-parameters` was mentioned in the doc for `activities-window-persistent-parameters`, but doesn't exist, so I removed that.  Vestigial?

> Also, forgive the nitpicking, but I'd suggest simplifying the `frame` binding one step further by making the lambda take the frame as its argument, and then passing `(selected-frame)` as the argument to the timer.

Sure, done.  In fact I went all the way and passed the copied state as an argument too.

alphapapa:
Thanks, merged.

PR #101, opened 2024-06-20
Bookmark fixes and features.

shipmints:
- Added missing bookmark-store to activities-new

- Added missing bookmark rename function to activities-rename

- Added delete bookmarks to activities-discard

- Removed superfluous calls to bookmark-maybe-load-default-file (both bookmark-store and bookmark-delete have had those calls in place internally for 30 years according to blame)

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -366,7 +366,9 @@ activity."
   (let ((activity (make-activities-activity :name name)))
     (activities-switch activity)
     (activities-set activity :state nil)
-    (activities-save activity :defaultp t)))
+    (activities-save activity :defaultp t)
+    (when activities-bookmark-store
+      (activities-bookmark-store activity))))
 
 (defun activities-rename (activity name)
   "Rename ACTIVITY to NAME."
@@ -375,10 +377,14 @@ activity."
           (name (read-string (format "Rename activity %S to: "
                                      (activities-activity-name activity)))))
      (list activity name)))
+  (when activities-bookmark-store
+    (activities-bookmark-delete activity))
   (setf activities-activities (map-delete activities-activities
                                           (activities-activity-name activity))
         (activities-activity-name activity) name
         (map-elt activities-activities name) activity)
+  (when activities-bookmark-store
+    (activities-bookmark-store activity))
   (activities--persist))
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


activities.el
```diff
@@ -375,10 +377,14 @@ activity."
           (name (read-string (format "Rename activity %S to: "
                                      (activities-activity-name activity)))))
      (list activity name)))
+  (when activities-bookmark-store
+    (activities-bookmark-delete activity))
```
alphapapa, line 381: Could this go in the other form added in this function?

activities.el
```diff
@@ -454,15 +460,17 @@ this demotes errors."
...
   (when (yes-or-no-p (format "Discard activity %S permanently?" (activities-activity-name activity)))
     (ignore-errors
       ;; FIXME: After fixing all the bugs, remove ignore-errors.
       (when (activities-activity-active-p activity)
         (activities-close activity)))
-    (setf activities-activities (map-delete activities-activities (activities-activity-name activity)))))
+    (setf activities-activities (map-delete activities-activities (activities-activity-name activity)))
+    (when activities-bookmark-store
+      (activities-bookmark-delete activity)))
+  )
```
alphapapa, line 473: Hanging paren.  :)

activities.el
```diff
@@ -881,13 +889,18 @@ with prefix argument, choose another activity."
...
          (bookmark-name (concat activities-bookmark-name-prefix activities-name))
          (props `((activities-name . ,activities-name)
                   (handler . activities-bookmark-handler))))
     (bookmark-store bookmark-name props nil)))
 
+(defun activities-bookmark-delete (activity)
+  "Delete the `bookmark' record for ACTIVITY."
+  (let* ((activities-name (activities-activity-name activity))
+         (bookmark-name (concat activities-bookmark-name-prefix activities-name)))
+    (bookmark-delete bookmark-name)))
```
alphapapa, line 902: It would be a bit more robust to find the bookmark to delete by finding ones that have the properties matching the ones made by the store function.  That way it will still work even if the bookmark has been renamed.

alphapapa (changes requested):
Thanks.  I'll plan to merge this for v0.8.

activities.el
```diff
@@ -881,13 +889,18 @@ with prefix argument, choose another activity."
 
 (defun activities-bookmark-store (activity)
   "Store a `bookmark' record for ACTIVITY."
-  (bookmark-maybe-load-default-file)
```
alphapapa, line 884: Not that I distrust what you said about this, but I feel like I added this because I found it necessary in another package of mine, so I'm not sure it should be removed here.  I wouldn't want to introduce a bug.

alphapapa (changes requested):
Forgot to mention this.

jdtsmith:
I just added #121 but then saw this; maybe it belongs here.  Storing again an existing bookmark doesn't seem to cause any problems.  Of course it would be good to compare the bookmark-activity list with the activity list and cull any which have gone missing.

shipmints:
I could rebase these, again, and consider merging #121 but it's been a while.

PR #103, opened 2024-06-20
Formalized the mode-line with a clickable menu to resume activities. …

shipmints:
…Added a main menu-bar menu with all major commands.

No option to suppress the menu.

No work done on custom faces for the mode line.

No option to customize the mode-line appearance.

No work done on a context menu (not sure it's needed).

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -476,21 +476,22 @@ It will not be recoverable."
 
 (defcustom activities-mode-idle-frequency 5
   "Automatically save activities when Emacs has been idle this many seconds."
-  :type 'natnum)
+  :type 'natnum) ; TODO: add a setter?
 
 (defvar activities-mode-map (make-sparse-keymap)
   "The mode keymap for `activities-mode'.")
 
 (defun activities-mode-line-format ()
-  (when-let ((cur (activities-current)))
-    (let ((cur-activity-title (concat " " (activities-name-for cur))))
+  "Activities mode line format."
+  (when-let ((current-activity (activities-current)))
+    (let ((current-activity-title (concat " " (activities-name-for current-activity))))
       `(:propertize
-        ,cur-activity-title
+        ,current-activity-title
         mouse-face mode-line-highlight
         help-echo
         ,(lambda (&rest _)
            (concat
-            (format "Current activity:%s\n" cur-activity-title)
+            (format "Current activity:%s\n" current-activity-title)
             "mouse-1: Display minor mode menu\n"
             "mouse-2: Show help for minor mode"))
         keymap
@@ -505,10 +506,9 @@ It will not be recoverable."
                          (describe-function 'activities-mode)))
            map)))))
 
-(defcustom activities-mode-line '(:eval (activities-mode-line-format))
+(defcustom activities-mode-line-lighter '(:eval (activities-mode-line-format))
   "Activities mode line definition."
   :type 'sexp
-  :group 'activities
   :risky t)
 
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


alphapapa:
I appreciate your wanting to contribute substantially to this package.  I don't recall having discussed copyright assignment yet, though.  https://github.com/alphapapa/activities.el#copyright-assignment  Have you done that?

shipmints:
Yes.

alphapapa:
Very well, but I am not allowed to take your word for it.  The Emacs maintainers require me to verify by checking with them.  What name and email address should I ask them about?

alphapapa:
> Yes.

I happened to just notice this post on emacs-devel by you: https://lists.gnu.org/archive/html/emacs-devel/2024-06/msg00691.html  You have asked to begin the process, but you have not completed it, and I cannot accept your contributions until the process has been completed, and until I receive confirmation from the FSF secretary or the Emacs maintainers.

shipmints:
Perfect. How will you get that confirmation or should I copy/paste their response to me?

alphapapa:
When you send the form to the FSF, you can ask them to cc me on the confirmation.  Otherwise, you can tell me that the process is completed, and then I can ask the maintainers to confirm.

shipmints:
I've asked them to cc you and pasted your email address. Either way, I'll let you know. Thank you for your patience.

activities.el
```diff
@@ -479,12 +479,47 @@ It will not be recoverable."
   "Automatically save activities when Emacs has been idle this many seconds."
   :type 'natnum)
 
+(defvar activities-mode-map (make-sparse-keymap)
+  "The mode keymap for `activities-mode'.")
```
alphapapa, line 483: Strictly speaking, I'm not sure if this is necessary, or actually correct, because it's not necessary to use `activities-mode` to use the various commands, so I'm not sure if the menu bar entries should be tied to a mode/map.

activities.el
```diff
@@ -479,12 +479,47 @@ It will not be recoverable."
...
+        ,(let ((map (make-sparse-keymap)))
+           (define-key map [mode-line down-mouse-1]
+                       activities-mode-line-menu)
+           (define-key map [mode-line down-mouse-3]
+                       activities-mode-line-menu)
+           (define-key map [mode-line mouse-2]
+                       (lambda ()
+                         (interactive)
+                         (describe-function 'activities-mode)))
+           map)))))
```
alphapapa, line 507: 1. Nitpick, but I prefer not to abbreviate variable names in most cases, e.g. `cur`.
2. Probably the keymap should be defined statically.
3. I guess the lambda could be defined statically as well, i.e. not be a closure over the title variable, because it should be able to get the current activity name by doing what this function does when necessary.
4. The function should have a docstring.  :)

activities.el
```diff
@@ -479,12 +479,47 @@ It will not be recoverable."
...
+
+(defcustom activities-mode-line '(:eval (activities-mode-line-format))
+  "Activities mode line definition."
+  :type 'sexp
+  :group 'activities
+  :risky t)
```
alphapapa, line 513: 1. I think this should be called `activities-mode-lighter`, unless I'm missing something.  
2. It shouldn't be necessary to specify the group, since it comes after the `defcustom` earlier in the file, right?

activities.el
```diff
@@ -479,12 +479,47 @@ It will not be recoverable."
...
+
 ;;;###autoload
 (define-minor-mode activities-mode
   "Automatically remember activities' state.
 accordingly."
   :global t
   :group 'activities
+  :lighter activities-mode-line
+  :keymap activities-mode-map
```
alphapapa, line 522: See earlier comment about menu bar and mode map.

Alternatively, we could consider defining this map as a prefix map for the suggested bindings.

activities.el
```diff
@@ -495,6 +530,59 @@ accordingly."
       (setf activities-mode-timer nil))
     (remove-hook 'kill-emacs-hook #'activities-mode--killing-emacs)))
 
+(require 'easymenu)
```
alphapapa, line 533: It would be good to add a header comment for this section.

activities.el
```diff
@@ -495,6 +530,59 @@ accordingly."
...
+                  (let ((current-activity-name
+                         (when-let ((current-activity (activities-current)))
+                           (activities-activity-name current-activity))))
+                    (mapcar (lambda (act)
+                              (vector act `(activities-resume (activities-named ,act))
+                                      :style 'radio
+                                      :selected (equal current-activity-name act)
+                                      ))
+                            (activities-names))))
+        ))
```
alphapapa, line 548: I'm not sure what this is about, but a bare `setq` on an undefined variable is bogus, and there are hanging parens.  :)

activities.el
```diff
@@ -495,6 +530,59 @@ accordingly."
...
+     :help "Focus on the specified buffer from a live activity (activities-tabs-mode only)"]
+    ["List" activities-list
+     :help "Show the master list of known activities"]
+    ["Rename" activities-rename
+     :help "Rename the specified activity"]
+    ["Discard" activities-discard
+     :help "Discard the specified activity (this is undoable)"]
+    ["Save All" activities-save-all
+     :help "Save all live activities to disk"]
+    ))
```
alphapapa, line 578: It would be helpful to roughly organize these entries by purpose, with some separators between major types.

Also, hanging parens.  :)

activities.el
```diff
@@ -495,6 +530,59 @@ accordingly."
...
+
+(easy-menu-define activities-mode-line-menu nil
+  "Activities Mode Line Menu"
+  `("Activities"
+    ,activities-menu-item-resume
+    ))
```
alphapapa, line 584: Also, hanging parens.  :)

alphapapa (changes requested):
Thanks.

activities.el
shipmints: We need the mode's keymap for the menu to appear on the menu bar. If the mode isn't active, the menu won't appear so no harm. It doesn't affect general operation.

activities.el
shipmints: Done. I left the closure and keymap as is since that seems to be the general approach via informal surveillance (and it works fine and without any overall performance issues).

activities.el
shipmints: Done.

activities.el
shipmints: Let's address prefix map and suggested bindings separately.

activities.el
shipmints: Done.

activities.el
shipmints: Done. Now a defvar.

activities.el
shipmints: Done.

activities.el
shipmints: Done.

shipmints:
FYI, I added a defensive test in the mode before setting the timer, should someone set activities-mode-idle-frequency to nil (as I might).

Issue #107, opened 2024-06-22
Need to display tabs correctly with tab-bar-format-tabs-groups under tab-bar-mode


shipmints:
```activities``` needs to respect ```tab-bar-mode```'s ```tab-bar-tab-group-face-function``` override in the same way it currently respects ```tab-bar-tab-face-function```. The lack of support means that the ```activities-tab``` face is never applied.

A PR coming shortly with the tested fix for this.

Configuration snippet for testing:
```elisp
  (setq tab-bar-format
        '(
          tab-bar-format-tabs-groups ;; this shows groups
          ;; tab-bar-format-tabs ;; switch to this for plain tabs
          tab-bar-separator
          activate
          tab-bar-format-align-right
          tab-bar-format-global
          ))
```
With groups enabled, you'll have to assign a group to one tab to get the group to appear and for the broken formatting path to be invoked.

You can invoke this to ensure there's at least one group for testing ```(tab-bar-change-tab-group "*default*")```

or call ```M-x tab-group``` to assign a name interactively.

P.S. While reading through the ```tab-bar-mode``` code, it appears there is a bug that, if fixed, might render the PR correction unneeded. Perhaps in Emacs 30 if this is truly a bug and gets reported.
```elisp
(defun tab-bar-tab-group-face-default (tab)
  (if (not (or (eq (car tab) 'current-tab)
               (funcall tab-bar-tab-group-function tab)))
      'tab-bar-tab-ungrouped
    (tab-bar-tab-face-default tab)))
```
At the end of the function, ```tab-bar-tab-face-default``` is called and not ```tab-bar-tab-face-function``` so the override is never applied when groups are being formatted. Since the override would have been set to ```activities-tabs--tab-bar-tab-face-function``` the custom face is never applied. Also, to get around this, one can't mix both the tab format and the group format or the tabs get repeated.

shipmints:
See #108.

alphapapa:
> P.S. While reading through the `tab-bar-mode` code, it appears there is a bug that, if fixed, might render the PR correction unneeded. Perhaps in Emacs 30 if this is truly a bug and gets reported.

Would you please report that bug, then?  If it is truly a bug (you should probably cc Juri on it, as he's the maintainer of that library), then it might not be too late to get it fixed in Emacs 30.  Then, if it's as you say, we might be able to omit the change here.

shipmints:
[...]

PR #108, opened 2024-06-22
Fixes for tab formatting when using tab-bar-format-tabs-groups in tab…

shipmints:
…-bar-format.

A parallel implementation to
activities-tabs-tab-bar-tab-face-function-original was added to support group tab formatting under the predictable name activities-tabs-tab-bar-tab-group-face-function-original.


alphapapa:
See https://github.com/alphapapa/activities.el/issues/107#issuecomment-2184284386; please ensure that this change is really needed before we make it.

shipmints:
Rebased.

PR #110, opened 2024-07-10
Change for #88 - activities-save-all adjusts height minibuffer

jeff-phil:
* Moved check of `activities-anti-save-predicates` in `activities-save` to be first
* Added `activities-anti-save-predicates` to `activities-save-all` for future-proof, e.g. in case `activities-with` is used in `activities-save-all` in future for some reason.
* Unset `activities-anti-save-predicates` during `activities-mode--killing-emacs` in order to ensure saving when killed.

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -434,13 +434,14 @@ With PERSISTP, persist to disk (otherwise see
 `activities-always-persist').  To be safe for `kill-emacs-hook',
 this demotes errors."
   (interactive)
-  (with-demoted-errors "activities-save-all: ERROR: %S"
-    (dolist (activity (cl-remove-if-not #'activities-activity-active-p (map-values activities-activities)))
-      (let ((activities-saving-p t)
-            ;; Don't write to disk for each activity.
-            (activities-always-persist nil))
-        (activities-save activity :lastp t)))
-    (activities--persist persistp)))
+  (unless (run-hook-with-args-until-success 'activities-anti-save-predicates)
+    (with-demoted-errors "activities-save-all: ERROR: %S"
+      (dolist (activity (cl-remove-if-not #'activities-activity-active-p (map-values activities-activities)))
+        (let ((activities-saving-p t)
+              ;; Don't write to disk for each activity.
+              (activities-always-persist nil))
+          (activities-save activity :lastp t)))
+      (activities--persist persistp))))
 
 (defun activities-revert (activity)
   "Reset ACTIVITY to its default state."
@@ -497,7 +498,8 @@ accordingly."
 (defun activities-mode--killing-emacs ()
   "Persist all activities' states.
 To be called from `kill-emacs-hook'."
-  (let ((activities-always-persist t))
+  (let ((activities-always-persist t)
+        (activities-anti-save-predicates nil))
     (activities-save-all)))
 
 ;;;; Functions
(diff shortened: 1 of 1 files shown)


jeff-phil:
@alphapapa 

Based on discussion for #88 and #89, this PR changes flow to check `activities-anti-save-predicates` earlier in `activities-save` and adds same to `activities-save-all` for future-proofing and efficiency.

I also changed `activities-mode--killing-emacs` to ignore `activities-anti-save-predicates` during killing of emacs, in order to ensure saving and not losing activities state.  Logic could first close minibuffer, and then save. And/or same with closing active backtrace. Let me know if want changes there.

Did branch of fork, instead of master->master.  ;)

Thanks as always for review and feedback!

activities.el
```diff
@@ -434,13 +434,14 @@ With PERSISTP, persist to disk (otherwise see
...
 this demotes errors."
   (interactive)
-  (with-demoted-errors "activities-save-all: ERROR: %S"
-    (dolist (activity (cl-remove-if-not #'activities-activity-active-p (map-values activities-activities)))
-      (let ((activities-saving-p t)
-            ;; Don't write to disk for each activity.
-            (activities-always-persist nil))
-        (activities-save activity :lastp t)))
-    (activities--persist persistp)))
+  (unless (run-hook-with-args-until-success 'activities-anti-save-predicates)
```
alphapapa, line 437: My idea was to have a new, global anti-save-predicates list that would prevent saving of any activities.  I think that the existing anti-save-predicates list should continue to exist as a per-activity check done after switching to an activity, before saving it; making that list global now would be a change in functionality that could break user configuration.  Sorry if I wasn't clear.

activities.el
```diff
@@ -497,7 +498,8 @@ accordingly."
 (defun activities-mode--killing-emacs ()
   "Persist all activities' states.
 To be called from `kill-emacs-hook'."
-  (let ((activities-always-persist t))
+  (let ((activities-always-persist t)
+        (activities-anti-save-predicates nil))
```
alphapapa, line 502: I'm not sure this is a good idea.  The idea of the anti-save-predicates is that, if any of them return non-nil, some condition is such that the activity should not be saved (e.g. it might mean that saving would fail, or that trying to restore the activity from that state would fail).  That Emacs is exiting is not necessarily a reason to ignore that condition.  I think we shouldn't do this unless we know a good reason to make this exception; and even then, it might be better to make it optional, or to suggest some way to do the check within each predicate.

alphapapa (changes requested):
Hi Jeff,

First, as I mentioned on #89, regarding the FSF copyright assignment, I'm not allowed to take anyone's word for it (not that I doubt you). I have to get confirmation from the Emacs maintainers (or find you in the list of existing contributors). Can I find commits by you in emacs.git? If so, under what name/email?

I'll submit some other feedback as review comments.  Thanks.

jeff-phil:
> First, as I mentioned on #89, regarding the FSF copyright assignment, I'm not allowed to take anyone's word for it (not that I doubt you). 

Oops, I missed that comment on #89.  Yes, understood and agree with your due diligence!

> I have to get confirmation from the Emacs maintainers (or find you in the list of existing contributors). Can I find commits by you in emacs.git? If so, under what name/email?

Nothing in emacs.git (yet!).  I've looked for the contributors list on Savannah, but never could find.  Wish they made it easier to verify.  :)

Anyway, here's gist with identity details: https://gist.github.com/jeff-phil/005fa719795af2a6a537ef16a4f6f00c.  In there you will notice link to my FSF copyright agreement (https://0x0.st/s/vfAx-oCHIbxjEdXj-iVukQ/XM_U.tar.gz ) which has public signatures for me and FSF to verify the digital signatures in the agreement.  My OpenPGP fingerprint may also be found here: https://keys.openpgp.org/search?q=jeffphil%40gmail.com 

If need me to have assign@fsf.org send confirmation, I can do that as well.  Just let me know if good, or know of a better way to verify!

I still need to digest the other review items you put here.  Not sure I understand exactly what you are saying by "new global anti-save-predicates list" but I'll noodle on it a bit more.

Thanks!

activities.el
jeff-phil: > My idea was to have a new, global anti-save-predicates list ...

Relaying what I think you mean:

- `activities-anti-save-predicates` is to be "activity" specific, meaning if there is an activity _**Foo**_ then there could be an `activities-anti-save-predicates` that would only apply (_be local_) to activity _**Foo**_.

- Add an `activities-global-anti-save-predicates` (or similarly named) variable that, given having 2 activities _**Foo**_ and _**Bar**_, would apply to both activities and prevent saving of both.

Am I on the right track for what you were thinking?  Based your response, I will probably have a few follow-up questions.

> making that list global now would be a change in functionality that could break user configuration

You do caveat in the README: **_Because `activities` is at an early stage of development..._** :)

activities.el
jeff-phil: I figured that would be a bit controversial, I will remove next iteration.

activities.el
alphapapa: What I mean is that:

1. `activities-anti-save-predicates` should be unchanged from how it works now.  When about to save an activity, after switching to it, and before saving it, these predicates are called to see whether the activity should be saved.
2. A new, similar list of predicates should be defined for globally preventing *any* saving of activities.  Its predicates should be called before switching to any activities, and if any of them return non-nil, there should be no attempt to save any activity, nor should any activity be switched to.

alphapapa:
Hi Jeff,

Thanks for your patience.  I looked at the FSF CA tarball you uploaded; FYI the PDF in it is unsigned and undated.  Anyway, all I need to know is the name and email address that you executed the CA under, and then I can ask the Emacs maintainers to confirm it for me.

jeff-phil:
Hi Adam-

I have been completely out-of-pocket for last several weeks, so no problem at all.

Yes, anymore with digital signatures (in US) don't need actual manual signatures and dates, since digital signatures contain both.

My FSF CA email is `jeffphil@gmail.com`.

Thanks!

jeff-phil:
I sent email with details.  Thanks!

PR #114, opened 2024-07-25
Fix: (activities-bookmark-handler) Add autoload cookie

josephmturner:
Previously, if the user ran bookmark-jump and selected an activity bookmark before activities.el was loaded, Emacs would signal an error.


alphapapa:
Thanks, Joseph!

PR #119, opened 2024-09-07
Xref history per activity

theyamo:
Xref history is kept for each activity. When user is not currently in activity and uses xref functionality, the original xref history function is called.

Currently, xref history is not persisted.

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -66,6 +66,7 @@
 (require 'map)
 (require 'persist)
 (require 'subr-x)
+(require 'xref)
 
 ;;;; Types
 
@@ -200,6 +201,12 @@ deserialized back to the buffer after it is reincarnated.")
 
 (defvar activities-saving-p nil
   "Non-nil when saving activities' states.")
+
+(defvar activities-xref-history-storage-function-original nil
+  "Original function for updating xref history, to be called when not in activity.")
+
+(defvar activities-xref-activity-history nil
+  "(ACTIVITY-NAME . (BACKWARD-STACK . FORWARD-STACK)) alist of Xref markers for each activity.")
 
 ;;;; Customization
 
@@ -488,10 +495,16 @@ accordingly."
       (progn
         (setf activities-mode-timer
               (run-with-idle-timer activities-mode-idle-frequency t #'activities-save-all))
+        (unless activities-xref-history-storage-function-original
+          (setq activities-xref-history-storage-function-original xref-history-storage))
+        (setq xref-history-storage 'activities--xref-history)
         (add-hook 'kill-emacs-hook #'activities-mode--killing-emacs))
     (when (timerp activities-mode-timer)
       (cancel-timer activities-mode-timer)
       (setf activities-mode-timer nil))
+    (when activities-xref-history-storage-function-original
+      (setq xref-history-storage activities-xref-history-storage-function-original)
+      (setq activities-xref-history-storage-function-original nil))
     (remove-hook 'kill-emacs-hook #'activities-mode--killing-emacs)))
 
 (defun activities-mode--killing-emacs ()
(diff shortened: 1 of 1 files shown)


alphapapa:
Hello,

This is a very interesting idea.  I think it would be very useful, indeed.

A few issues before it could be merged:

1. Please see this note in the readme: https://github.com/alphapapa/activities.el?tab=readme-ov-file#copyright-assignment
2. While this code is quite simple, I think this should be implemented as a separate minor mode so users could choose whether to enable it.  One of the principles I'm developing Activities by is to be flexible and integrate with other features optionally, rather than forcing users to modify their existing workflows.
3. One minor note: I think the xref history map can use the `activity` struct directly as the key rather than the name of the activity.  As is, I see no reason not to do so.  What do you think?

Thanks for submitting this patch.

activities.el
```diff
@@ -901,6 +914,24 @@ ignored."
...
+If not in activity, call the original xref function.
+
+Override existing value with NEW-VALUE if NEW-VALUE is set."
+  (let ((current-activity (activities-current))
+        (activity-name (when (activities-current)
+                         (activities-activity-name (activities-current)))))
+    (if current-activity
+        (progn
+          (unless (map-contains-key activities-xref-activity-history activity-name)
+            (map-put activities-xref-activity-history activity-name (xref--make-xref-history)))
```
alphapapa, line 929: Also, please note the docstring of `map-elt`: the best way now is to use `(setf (map-elt ...) VALUE)`.

activities.el
```diff
@@ -901,6 +914,24 @@ ignored."
...
+          (when new-value
+            (map-put activities-xref-activity-history activity-name new-value))
```
alphapapa, line 931: Same as above.

alphapapa (changes requested):
A couple more things.  :)

theyamo:
It occurred me right away to add this feature when I started using the activities package, because the two modes of operation that emacs offers for xref history are, IMO, either too coarse (global history), or too narrow (window local history).

Emacs hasn't had a concept similar to session/activity to establish a more balanced way of working in between of those extremities.

I used the activity name as the list key for development purposes. The struct itself as the key made the xref storage variable hard to read. It's removed now. I also did the other changes as requested.

I have done the paperwork already some years ago.

alphapapa:
> It occurred me right away to add this feature when I started using the activities package, because the two modes of operation that emacs offers for xref history are, IMO, either too coarse (global history), or too narrow (window local history).
> 
> Emacs hasn't had a concept similar to session/activity to establish a more balanced way of working in between of those extremities.

Indeed, I hope Activities can serve that purpose well, to optionally integrate and provide context for various features like that.

> I used the activity name as the list key for development purposes. The struct itself as the key made the xref storage variable hard to read.

Yeah, that can be a problem with Edebug and the like.  Setting various `print-level` variables can help, but sometimes even that isn't quite enough.

>  It's removed now. I also did the other changes as requested.

Thanks.
> 
> I have done the paperwork already some years ago.

That's good, but the Emacs maintainers require me to verify new (to me) contributors' status through them before accepting such contributions.  Forgive me if I should know who you are already.  :)  Please let me know the name and email address you completed the FSF CA form with, and I'll ask them for verification.

theyamo:
My name and email address have not changed. Hopefully you can get their confirmation.

alphapapa:
> My name and email address have not changed. Hopefully you can get their confirmation.

I do not know your name or email address.

theyamo:
It's visible in the git commit headers.

theyamo:
FWIW, i tried to email fsf.org about this myself (check if my documents are valid for these contributions) but didn't get a reply. Maybe I didn't ask it clearly enough.

alphapapa:
> They're visible in the git commit headers.

Since they aren't visible on GitHub's UI, I was hoping you'd be kind enough to save me the extra steps of having to view the PR's commits directly.  As well, the name and address configured in git does not necessarily match the ones a contributor used when completing FSF CA paperwork, so it's best to just state them for clarity.  It's generally helpful to make the maintainer's job as easy as possible, because there's never enough time to do all the things, especially when multiple projects are on one's plate.

theyamo:
Apologies.

I confirmed with FSF that my gnu paperwork is still valid.

alphapapa:
> Apologies.
> 
> I confirmed with FSF that my gnu paperwork is still valid.

As I said earlier, the Emacs maintainers require me to get confirmation from them; I am not allowed to take your word for it.

alphapapa:
I received your email with name and address, and I cc'ed you on my email to the maintainers.

alphapapa:
Eli confirmed your CA.  Thanks for your patience.  Haven't had much time for Emacs things lately.  Will merge this for the next version.

PR #121, opened 2024-10-25
Store activities on save

jdtsmith:
Store each activity whenever it is saved.  This handles cases in which
`activities-bookmark-store` was turned on after an activity was
defined, or the bookmark was deleted.


alphapapa:
Hi JD,

Yeah, I guess this is the right thing to do.  My only hesitation is that it may cause the bookmark file to be saved more frequently, which is mostly only a potential problem in light of the message showing up in the minibuffer, haha.  

What do you think?  Thanks.

jdtsmith:
Added a check for an existing bookmark with that name, so we save only if it doesn't yet exist.

alphapapa:
Thanks.  Does that need to be an option, or could we make that behavior unconditional?

jdtsmith:
Sorry just saw this.  In the current setup, when creating a new activity (via `activities-define`), you can _re-use_ an old bookmark name and it will overwrite the (vestigial) bookmark with new properties immediately.  This seems like a good thing to have, especially until bookmark removal on activity deletion is implemented.

Issue #130, opened 2025-01-10
Correct way to handle buffers opened with TRAMP? Particularly when the buffers and underlying host are no longer available.


trharmon0:
I saved a couple of Activities representing coding work on remote systems with buffers opened over TRAMP. The underlying host (an its buffers) are no longer available. Now, even with a TRAMP buffer and session cleanup via TRAMP's interactive util functions, activities-resume and a couple of other activities functions hang indefinitely and require a C-g to get Emacs back under control.

Curious if you have any advice on how this should be handled? It doesn't necessarily seem like a problem with the activities library so much as the usual silliness from TRAMP. But perhaps there's something we could do to recover from the activities side?

alphapapa:
Hi,

Hm, that's a bit awkward, indeed.  My first thought is that TRAMP is not timing out quickly enough, but the problem may not even be that simple, because I'm reminded of occasional issues I've noticed in which I have a TRAMP dired or file-backed buffer open that's pointing to an SFTP server, and when I call a command to switch to another buffer, and my completion framework wants to display metadata about it, it hangs while waiting for TRAMP to reestablish a connection to the server, which is quite annoying.

Anyway, if the TRAMP-backed buffers are in the activity's last-saved state, it seems to me that resetting the activity should remove such buffers from the activity, so resuming the activity shouldn't try to recreate the TRAMP-backed buffers.  If the TRAMP-backed buffers are in the activity's default state, you probably need to kill the TRAMP-backed buffers and redefine the activity (`C-u C-x C-a C-d`) without them.

Please try those ideas and let me know if it helps.  If the problem persists, we might need to deal with the windows' recent buffers lists, or something like that.

trharmon0:
These "zombie" activities continue to cause issues. I'm unsure how to kill/clean/reset them. I'm unable to dired-style mark and kill them via activities-list. If I try to create a new one with the same name as an old problematic one that is not allowed. If I try to resume a known problematic one I can wait until the TRAMP timeout expires to get into the activities but then the activities-kill hangs indefinitely.

I thought perhaps I could backdoor the process by going straight to the bookmarks backend and editing there but the problematic ones are not shown via list-bookmarks.

Thoughts?

trharmon0:
[...]

PR #133, opened 2025-01-23
annotate: skip file-truename for remote files

jdtsmith:
The test for a "dirty" activity used by the annotation function and
when updating the timestamp within the activity is based on whether
the "set of buffers and files" has changed.  We use `file-truename` to
look for "real" file changes but that can hang on remote files.  To
alleviate, we only check the `file-truename` on non-remote files.

### DIFF
--- a/README.org
+++ b/README.org
@@ -170,6 +170,9 @@ The sort order when completing activities can be configured using ~activities-so
 *Changes*
 + During save, the ~time~ slot for an activity remains unchanged unless its buffers or files differ from their last saved state.  ([[https://github.com/alphapapa/activity.el/pull/83][#83]].  Thanks to [[https://github.com/jdtsmith][JD Smith]].)
 
+*Fixes*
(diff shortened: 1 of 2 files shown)


jdtsmith:
This is really kind of a tramp issue, but if you save an activity with a remote file included that is accessible only in some locations (or some networks), `activities--buffers-and-files-differ` (from the new annotations) hangs.   TRAMP does this sort of thing all the time.  This is used to compute the "dirty flag" (`*`) on annotations, and as well to update the timestamp only when "buffer and files" change.   I think the right fix is the following: if the specified file is remote, simply return it instead of `(file-true name file)`.

alphapapa:
Ah, this sounds like it might be the solution for #130!

jdtsmith:
Oh interesting, hadn't seen that one.  Tramp sets some pretty deep hooks...

alphapapa:
Thanks for your patience.

Issue #109, opened 2024-07-02
Lost all my activities


julian-hoch:
Today I lost all my saved activities.
After deleting my elpa directory (due to some package conflicts) and reinstalling the packages, all activities are now gone.
I can still see them in the bookmarks, but they don't load anymore.
I was led to believe from the README that activities are stored in the bookmarks, but seems this is not actually a the case - instead, they are stored in a separate file in the "persist" subdirectory. This was not made clear in the documentation, so I did not put this file under version control (hopefully I can still find a backup somewhere).
In any case, I am not sure why reinstalling the package would delete this file. I am considering this a defect or at least a shortcoming in the documentation, and hope this is a warning to other users to backup this file.

alphapapa:
Hi Julian,

Sorry to hear about that.  I only ever had issues with losing the value of `activities-activities` when I was doing things during development that users shouldn't do, like calling `unload-feature` and then reloading a new version of the package; and that seemed to be some kind of weird interaction with the `persist` library.  Since I stopped doing that, I've never lost my activities list.

Without an exact list of the steps you followed, it would be hard to know what happened and why.  Since `persist` is not a built-in package, and since I don't know whether or in what order you might have restarted Emacs during those processes, it's hard to say what the various libraries might have done being initialized with various state.

FWIW, I tested the following:

1.  Using `with-emacs.sh`, install `activities` (and its dependencies).
2. Define a test activity.
3. Exit Emacs.
4. Delete its `elpa/` directory.
5. Restart Emacs.
6. Install `activities` again (including `persist`).
7. `M-x activities-resume RET`.

The result was that the `persist/activities-activities` file was still present and its value was still loaded.

The bookmark support is optional, and they only store a reference to an activity's name.  An activity data structure can be quite large, and in the past there have been occasional issues with encoding, so it didn't seem like a good idea to store them directly in the bookmarks file (wouldn't want to break that file).

In the meantime, we can certainly mention the location of the data file in the documentation if you think that would be helpful.  (I doubt most users would notice, but you're not most users.)

We could also allow the user to set the location of the data file, as `persist` seems to support that.  That could make it easier to back it up, but if this happened due to a bug in Activities, Persist, or some kind of interaction, I don't think moving the location of the file would fix it.

I'll loop in the Persist maintainer: @josephmturner I don't know if there's anything that can be done about this, but what do you think?

julian-hoch:
[...]

PR #147, opened 2025-10-04
Fix: Resolve data loss issue by upgrading persist.el to version 0.7

josephmturner:
See #109.


alphapapa:
Hi Joseph,

Thanks.  This seems like a good idea in general, but since it's not strictly necessary (i.e. we're not depending on API or behavioral changes in `persist`), I'm not sure if I should force users to upgrade it.  This library is included in other downstream software distributions, and doing this might preclude upgrades to this library from being distributed to users if their distribution doesn't also upgrade `persist` at the same time (which might be a good idea to do, but they might decide not to; e.g. Debian's policies are restrictive in this regard).

I think I might prefer that we add a warning if the detected version of `persist` is less than 0.8.  What do you think?

josephmturner:
Then again, I'd guess that there are more users installing Emacs packages with `package.el` than there are users installing packages using slow LTS package managers like Debian.  IIUC, bumping the version would potentially save some `package.el` users from this bug.  Right?

Oh and...  I'm not sure if Guix explicitly pulls in the version of `persist.el` specified in the header.

josephmturner:
I think my last point about Guix is bogus.  AFAICT, Emacs package definitions in Guix specify dependent Emacs packages as either `inputs` or `propagated-inputs` without specifying a version.  In this case, I think means that users will end up with the latest available version of `emacs-persist`.

I don't feel strongly about this issue.  Feel free to close or merge as you see fit :)

josephmturner:
Sounds good.  The percentage of folks who would be forced to upgrade `persist.el` is indeed small.  Thanks for explaining.

PR #149, opened 2025-10-08
Add convenience functions for resume by name and revert current

matsl:
# What

Add convenience functions for resume by name and revert current.

# Why

Add functions activities-resume-named and
activities-revert-current. These small functions makes it easier for a
user to create action buttons in Hyperbole like this:

    <activities-resume-named "Activity">
    <activities-revert-current>

Feedback from the functions use the name of the activity rather than
the activity itself.

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -416,6 +416,15 @@ available."
     (when (or resetp (not already-active-p))
       (activities-set activity :state (if resetp 'default 'last)))))
 
+(defun activities-resume-named (name &optional reset)
+  "Resume an activity given by NAME.
+If non-nil RESET roll back to activity's default configuration."
+  (let ((activity (activities-named name)))
+    (unless activity
+      (user-error "No activity: %s" name))
+    (activities-resume activity :resetp reset)
+    (message "Activity: %s." name)))
+
 (defun activities-switch (activity)
   "Switch to ACTIVITY.
 Interactively, offers active activities."
@@ -471,6 +480,12 @@ this demotes errors."
     (user-error "No active activity"))
   ;; TODO: Consider resetting the activity's buffers list.
   (activities-set activity :state 'default))
+
+(defun activities-revert-current ()
+  "Revert current activity to its default configuration."
+  (let ((activity (activities-current)))
+    (activities-revert activity)
+    (message "Reverted: '%s'." (activities-name-for activity))))
 
 (defun activities-discard (activity)
   "Discard ACTIVITY and its state.


alphapapa:
Hi Mats,

Thanks for submitting this PR.  These are some interesting ideas.

Regarding the `resume-named` command: I'm reluctant to add that, because it would be inconsistent with the existing API.  If I added that, it seems like I'd need to add `-named` versions of all the other commands, which would bloat the API and be confusing to users.

We have the `activities-named` function, so this boils down to `(activities-resume (activities-named NAME))`.  If you need to have that in a single function, I think you could safely add it to your own library, as I don't intend to change that API and cause breakage.

Alternatively, we could consider modifying all of the commands that accept an activity as its argument to also accept an activity name, so you could call `(activities-resume NAME-OR-ACTIVITY)`.  That would seem like a general improvement, although I'm still a bit reluctant, because it adds complexity to the API; and for the interactive case, `activities-completing-read` already handles looking up activities by name.  But I could probably be more easily convinced about this than the other solution.  :)

Regarding the `revert-current` command: I'm reluctant to add that for similar reasons: it seems like I'd need to add `-current` versions of all the existing commands.  In this case, `activities-revert` already calls `activities-current` in the interactive form.  So for the use case you cite, I'd recommend implementing a way to call interactive commands, if one doesn't already exist (i.e. what if you do `<call-interactively 'activities-revert>`?).  And as with the name issue, I don't intend to change the API, so if you need to add a local command that calls that, I wouldn't expect to break anytime soon.

So, respectfully, I don't want to make these changes in the project, as they would seem like special cases that add complexity and inconsistency.  But I'm happy to consider more general improvements that would make the entire library easier to use.

matsl:
Hi Adam, just a quick note:  Using <call-interactively 'activities-revert> has the downside that the activity struct is echoed in the message area, which IMHO is not that great. So although the call-interactively trick works it does not provide an ideal end result.

rswgnu:
> Adam wrote: "Alternatively, as we covered earlier, we could allow the commands to accept either a struct or a name as their argument, which I'm not strictly opposed to (it's not uncommon for Emacs commands or functions to accept multiple argument types), but you seemed to think that was a bad idea."

@alphapapa, yes, let's do this; Mats will send you a new PR with this change for yo uto evaluate.  It will also make the activity argument to activities-revert optional, defaulting to (activities-current) so without an argument, you can revert to the stored version of the activity.  The combination of these will solve all the problems we have.  Thanks.

PR #152, opened 2025-11-02
Make revert and resume callable as action buttons

matsl:
# What

Let activities-resume take a string argument so it can be called
directly using an activity's name.

Similarly let activities-revert activity parameter be optional and
default to current activity.

# Why

Proposing a change within the current API instead of adding new
functions. See PR #149 .

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -411,6 +411,10 @@ available."
   (interactive
    (list (activities-completing-read :prompt "Resume activity" :default nil)
          :resetp current-prefix-arg))
+  (when (stringp activity)
+    (unless (activities-named activity)
+      (user-error "No activity: %s" activity))
+    (setq activity (activities-named activity)))
   (let ((already-active-p (activities-activity-active-p activity)))
     (activities--switch activity)
     (when (or resetp (not already-active-p))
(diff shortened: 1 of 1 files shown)


activities.el
```diff
@@ -411,6 +411,10 @@ available."
   (interactive
    (list (activities-completing-read :prompt "Resume activity" :default nil)
          :resetp current-prefix-arg))
+  (when (stringp activity)
+    (unless (activities-named activity)
+      (user-error "No activity: %s" activity))
+    (setq activity (activities-named activity)))
```
rswgnu:
```emacs-lisp
  ;; Simplified change version:
  (when (stringp activity)
    (unless (setq activity (activities-named activity))
      (user-error "No activity: %s" activity)))
```

rswgnu (changes requested):

rswgnu (approved):

alphapapa:
Hi Mats,

Having thought about this for a while, I'd prefer not to modify Activities' APIs like this.  The API is consistent in taking an activity struct as the first argument to functions, and changing that would require making changes all over the codebase.  And I'd prefer not to introduce inconsistencies by only changing a few functions in that way, because that would also be confusing to users of the APIs, and it would add a lot of boilerplate-type code for little benefit.

You may be able to work around some of this by using advice in Hyperbole, for example:

```elisp
(with-eval-after-load 'activities
  (define-advice activities-revert (:filter-args (args) hyperbole--activities-revert-advice)
    (unless (car args)
      (setf (car args) (or (activities-current)
                           (user-error "No current activity"))))
    args))
```

However, you should be careful, as Activities itself applies some advice to some of its functions (see `activities-tabs-mode`), and the order in which the advice is applied and unapplied does matter.

rswgnu:
Mats and I will take care of this in Hyperbole then, integrating with activities.el as needed.  Thanks.

alphapapa:
Thanks, Bob.  I appreciate your wanting to integrate with Activities.  Please let me know if I can help in other ways.

PR #154, opened 2025-11-12
Give one more space for the annotation for minibuffer right align

manphiz:
Currently the annotation is perfectly right aligned, but this gives unpleasant display under console, which will force a line-wrap and push the ending `*' to the next line.  For example:

```
some-activitiy-1       1 bufs, 0 files, 2 days*
some-activitiy-2       2 bufs, 0 files, 3 days*
```
Will become
```
some-activitiy-1       1 bufs, 0 files, 2 days\
*
some-activitiy-2       2 bufs, 0 files, 3 days\
*
```
This commit adds one more space so that it looks better when running Emacs under console.

### DIFF
--- a/activities.el
+++ b/activities.el
@@ -984,7 +984,8 @@ which see, with DEFAULT."
                                      'face `(:foreground ,age-color :weight bold)))
                     (dirty-annotation (if dirtyp (propertize "*" 'face 'bold) " ")))
                  (concat (propertize " " 'display
-                                     `(space :align-to (- right ,(+ 1 (length annotation)
+                                     `(space :align-to (- right ,(+ (if (display-graphic-p) 1 2)
+                                                                    (length annotation)
                                                                     (length age-annotation)))))
                          annotation age-annotation dirty-annotation)))))
          (activity-table (str pred action)


alphapapa:
This change might solve the problem on terminal frames, but it will have a negative effect on GUI frames, right?

manphiz:
It will add one extra space for the right alignment on GUI, yes, and it will be consistent to what it looks like under CLI.

I wouldn't call it a "negative" effect as I don't consider an extra space on the right an eyesore.  But it would be good to hear from @jdtsmith on what he thinks of this change.

alphapapa:
If we have to add the `display-graphic-p` test, that's not a huge deal.  But we can't be the first ones to encounter this problem, as there are many other packages that display right-aligned text in the minibuffer, like Vertico/Marginalia/etc.  It would be good to check how they handle it; there are subtleties here, and something minor but important might have been overlooked (I've done that many times).

manphiz:
FYI I have updated the patch as suggested to keep the display in GUI unchanged.

jdtsmith:
It seems `marginalia` + `vertico`  face this same problem:

```elisp
(package-initialize)
(require 'marginalia)
(require 'vertico)
(vertico-mode)
(marginalia-mode)
(setq marginalia-align 'right)
```

then `C-x C-f`, leads to:

image

Note the truncation character `$` overwriting the last char in the date field.  It appears `vertico` turns on `truncate-lines` locally in the minibuffer.  This smells more like a bug in emacs' specified space calculation when you have no fringe (TTY) and need to consume one column for a wrap or truncation character.  I have not found an option to disable display of either.   

BTW, if you disable the fringe in GUI with `(set-fringe-style 0)` you get the same problem with `activities` or `vertico` + `marginalia:`

image

So `display-graphic-p` is an insufficient test.

Copying @minad in case he has any thoughts on right-alignment via specified space in TTY emacs where a wrap/truncation character is required.

minad:
> Note the truncation character $ overwriting the last char in the date field. It appears vertico turns on truncate-lines locally in the minibuffer. This smells more like a bug in emacs' specified space calculation when you have no fringe (TTY) and need to consume one column for a wrap or truncation character. I have not found an option to disable display of either.

See https://github.com/minad/vertico/pull/628#issuecomment-3229741786 for a theory.

One can overwrite the $ with a space to "disable" it, but the space will still be there. On GUI the solution is of course a fringe, even a 1px fringe suffices.

~~~ elisp
(set-display-table-slot standard-display-table 'truncation ?\s)
~~~

jdtsmith:
> One can overwrite the $ with a space to "disable" it, but the space will still be there

Maybe that's the right solution.  It feels like `:align-to (- right ,width)` should not require a truncation char if the text occupies the full width precisely.  Or that it should know that it will insist on one, and insert one fewer specified spaces.

manphiz:
> > One can overwrite the $ with a space to "disable" it, but the space will still be there
> 
> Maybe that's the right solution. It feels like `:align-to (- right ,width)` should not require a truncation char if the text occupies the full width precisely. Or that it should know that it will insist on one, and insert one fewer specified spaces.

I think you suggestion was to emulate the appearance by enabling "truncate-lines" and replace the truncation character with "`*`".  It does sound promising, so I tried to put those in my init.el to see how it looks.  It almost works well, except that the display engine seem to have some issue that is causing the last character to be missing, and can be dragged to more lines when scrolling (see the bottom right corner with the empty space instead of "*" in the screenshot.)

Screenshot_20251125_154415

This is happening both on Konsole and iTerm2 under macOS.

The proposed patch works around this issue obviously because it's not using the last character.  It's a bit ugly, yes, but seems to provide a more consistent display and fast.  Meanwhile for the emulation we need to carefully restore the user setting if they prefer not to truncate-lines, could be slower due to the mode toggle gymnastics, and still being affected by the display engine limitations.

Wdyt?

PR #156, opened 2026-02-07
Add transient menu for activities

lewang:
## Summary

Add `activities-transient.el`, a transient-based menu for common activities operations, and document
it in the README as an alternative to individual keybindings.

Groups commands into logical sections:
- **Create & Define**: new, define (with C-u to redefine default state)
- **Switch & Resume**: resume (with C-u to reset), switch active, switch buffer
- **Modify**: revert, rename
- **Close**: suspend, kill, discard
- **Other**: save all, list

## Usage

```elisp
(use-package activities-transient
  :after activities
  :bind
  ("C-x C-a" . activities-transient))
```

## Notes

- Only requires `transient` — no dependency on activities.el internals
- Uses autoload cookie for lazy loading
- README updated with transient menu configuration section

### DIFF
--- a/README.org
+++ b/README.org
@@ -76,6 +76,19 @@ This is the recommended configuration, in terms of a ~use-package~ form to be pl
      ("C-x C-a l" . activities-list)))
 #+END_SRC
 
+** Transient menu
+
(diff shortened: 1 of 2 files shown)


lewang:
I could make it a repo of itself, but seems more discover able for users in the same repo.

